PR #1, opened 2020-10-28
Add Pending Task Protocol proposal

### LINKED ISSUE #3: Pending State API
Hi @justinfagnani, you could also move the pending state proposal from your private repo here, we use a modified version of your proposal a lot and are quite happy with it.

edit: sorry, i missed the PR :)

justinfagnani:
This adds a Pending Task Protocol proposal, which enables elements to communicate the status of outstanding async work up the DOM tree.

Closes #3

### DIFF
--- a/proposals/pending-state.md
+++ b/proposals/pending-state.md
@@ -1,42 +1,47 @@
-# pending-state-protocol
+# pending-task-protocol
+
 An open protocol for interoperable asynchronous Web Components.
+
+Author: Justin Fagnani
 
 Document status: Draft
 
+Last update: 2020-10-28
+
 # Background
 
-There are a number of scenarios where a web component might depend on or perform some asynchronous work. Components may lazy-load parts of their implementation or content, perform I/O in response to user input, manage long-running computation, etc.
+There are a number of scenarios where a web component might depend on or perform some asynchronous task. Components may lazy-load parts of their implementation or content, perform I/O in response to user input, manage long-running computation, etc.
 
-It's often desirable to communicate the state of async work up the component tree to parent components so that they can display user affordances indicating whether the UI or content is pending some async operation.
+It's often desirable to communicate the state of async tasks up the component tree to parent components so that they can display user affordances indicating whether the UI or content is pending some async operation.
 
 Frameworks have the ability to invent custom APIs and patterns to handle this kind of cross-component communication. To enable the same with web components in a way that's interoperable between components from diffrent sources, and possibly implemented with different libraries, we need a protocol that components can implement without reliance on a common implementation.
 
 # Goals
 
-* Allow components to communicate that they have pending asynchonous work to ancestors in the DOM tree.
-* Allow components to know if there is pending asynchronous work in descendants in the DOM tree.
-* Allow components to intercept, modify, and block pending state notifications.
+* Allow components to communicate that they have pending asynchonous tasks to ancestors in the DOM tree.
+* Allow components to know if there are pending asynchronous tasks in descendants in the DOM tree.
+* Allow components to intercept, modify, and block pending task notifications.
 * Give guidance on the type of asynchronous work that should be notified.
 * Allow for some labelling or differentiation between types of asynchronous work.
 
 # Details
 
-## Asynchronous States
+## Asynchronous Task States
 
-There are four main states that an asynchronous operation can be in:
+There are four main states that an asynchronous task can be in:
 
 * Not-started
 * Started
 * Sucessfully completed
 * Failed
 
-## The `pending-state` event
+## The `pending-task` event
 
-Components with pending work indicate so by firing a composed, bubbling, `pending-state` Event, with a `promise` property:
+Components with pending tasks indicate so by firing a composed, bubbling, `pending-task` Event, with a `promise` property:
 
 TypeScript interface:
 ```ts
-interface PendingStateEvent extends Event {
+interface PendingTaskEvent extends Event {
   promise: Promise<void>;
 }
 ```
@@ -44,9 +49,9 @@ interface PendingStateEvent extends Event {
 Example:
 
 ```js
-class PendingStateEvent extends Event {
+class PendingTaskEvent extends Event {
   constructor(promise) {
-    super('pending-state', {bubbles: true, composed: true});
+    super('pending-task', {bubbles: true, composed: true});
     this.promise = promise;
   }
 }
@@ -56,7 +61,7 @@ class DoWorkElement extends HTMLElement {
   async doWork() { /* ... */ }
 
   startWork() {
-    this.dispatchEvent(new PendingStateEvent(this.doWork()));
+    this.dispatchEvent(new PendingTaskEvent(this.doWork()));
   }
 }
 
@@ -64,7 +69,7 @@ class DoWorkElement extends HTMLElement {
 class IndicateWorkElement extends HTMLElement {
   constructor() {
     super();
-    this.addEventListener('pending-state', async (e) => {
+    this.addEventListener('pending-task', async (e) => {
       this.showSpinner();
       await e.promise;
       this.hideSpinner();
@@ -73,12 +78,12 @@ class IndicateWorkElement extends HTMLElement {
 }
 ```
 
-The Promise must be resolved when the work is complete and rejected if the work fails. The value the Promise resolves to is unspecified.
+The Promise must be resolved when the task is complete and rejected if the task fails. The value the Promise resolves to is unspecified.
 
 Using an event with a Promise allows us to represent three of the four asynchronous states:
 
 * Not-started: not represented
-* Started: `pending-state` event fired
+* Started: `pending-task` event fired
 * Sucessfully completed: Promise resolved
 * Failed: Promise rejected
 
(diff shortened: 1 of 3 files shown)


proposals/pending-state.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Open Questions
+
+## Should the event carry a Promise or should there be two events?
+
+Animations have multiple events: `animationstart` and `animationend`. Promises make it very easy to correlate the start of an operation with the end of that operation.
+
+## Event Name
+
+Is `pending-state` good? Should it be `pending-work`?
```
daKmoR, line 93: `pending-work` sounds better to me 🤗

proposals/pending-state.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Types of Async Work
+
+There are some types of async work that we may not want to display UI affordances (like progress indicators) for. Async rendering used in order to yield to the browser's task queue for input hanlding and layout/paint for instance, should probably not trigger a progress indicator. Yet, code that measures style or layout may need to wait for rendering to complete.
+
+We could add a field to the event that indicates the type of work and standardize a small number of types, such as `loading` and `rendering`. An event to indicate async rendering starting and stopping has existing analogies with the `animationstart` and `animationend` events.
```
daKmoR, line 99: passing on the type seems like a good idea 👍

would enable use cases like this measuring render timings and showing a spinner at the same time 👌
```js
this.addEventListener('pending-work', async (ev) => {
  if (ev.workType === 'loading') {
    this.showSpinner();
    await ev.promise;
    this.hideSpinner();
  } else if (ev.workType === 'rendering') {
    this.startRenderingMeasure();
    await ev.promise;
    this.stopRenderingMeasure();
  }
}
```

proposals/pending-state.md
DeadWisdom: +1 pending-work -- "state" has connotations of redux and similar systems.

proposals/pending-state.md
```diff
@@ -0,0 +1,105 @@
...
+```ts
+interface PendingStateEvent extends Event {
+  promise: Promise<void>;
+}
+```
+
+Example:
+
+```js
+class PendingStateEvent extends Event {
```
Westbrook, line 47: Extending `event` gives us the full range of the class to customize data passing, and it looks like the proposal gets into that by hanging `promise` directly thereon and then postulating the inclusion of `workType` and possibly `workProgress` (or the like) as further additions. As this is in someway an early template of future event centric protocols is there any additional shape that we'd like to posit re: "What a protocol event _looks_ like"? This could help to both define these events as "Protocol events" and reduce the cognitive load of adopting/processing additional "Protocol events" after a developer has been introduced to the concept as a whole.

Prior art that I can think of likely goes way too far in this direction, but think of how the presence of `detail` clearly sets `CustomEvents` apart from `Events`. Again, extending `Event` gives up far more flexibility and power but I do believe some form of structure (even if it's just the suggestion of structure) will really benefit when this concepts moves beyond the first couple of protocols.

proposals/pending-state.md
```diff
@@ -0,0 +1,105 @@
...
+There are four main states that an asynchronous operation can be in:
+
+* Not-started
+* Started
+* Sucessfully completed
+* Failed
+
+## The `pending-state` event
+
+Components with pending work indicate so by firing a composed, bubbling, `pending-state` Event, with a `promise` property:
```
Westbrook, line 35: Do you see `cancelable` as having any benefit to this sort of protocol? This is how I'd expect to `Allow components to intercept, modify, and block pending state notifications.` with native events. The following could be of interest here:
```
const keepDoingWork = this.dispatchEvent(new PendingStateEvent(this.doWork()));
if (!keepDoingWork) {
    this.abortPendingWork();
}
```
This could similarly be achieved (as well as non-sync work cancelling) by outlining where to hang an `AbortController` off of the Protocol event.

proposals/pending-state.md
Westbrook: Kabob casing this may align with custom element naming and its goal not to step on native things, but what if `pending` was used to further generalize this events relationship to `work`, `loading`, `rendering`, `progress`, etc?

justinfagnani: `pending-task` is another idea.

I'm a little way of such a short name like `pending`, just for increased collision chances.

proposals/pending-state.md
justinfagnani: I _suspect_ a type field is generally a good idea, but I'd want to get a lot of feedback from app builders on what types they'd actually use.

proposals/pending-state.md
justinfagnani: I don't think there needs to be any standard for a "protocol event". They're just Events that carry whatever data they need for their specific task. Just like the platform events have different fields based on their class, so can our events. 

The big difference with events described here is that we'll only described the interface, not provide a common implementation.

proposals/pending-state.md
justinfagnani: That's an interesting idea. Do you know of specific use-cases for this? Do analogous framework features, like Suspense, allow this?

proposals/pending-state.md
justinfagnani: I renamed to `pending-task` for now. I think it's slightly better than `pending-work` and it felt right to replace instance of both "state" and "work" in the docs with "task".

proposals/pending-state.md
justinfagnani: Thinking about this some more, the AbortController pattern works when the initiator of the task creates and passes along the controller. Receivers of the event are not the initiators and don't necessarily have permission to abort.

Any event dispatcher can look at `event.defaultPrevented` if they want a synchronous signal to not perform work. Asynchronous cancellation would require the event to carry a cancel callback.

proposals/pending-state.md
peschee: @justinfagnani we currently use a variation of your implementation of pending-state container (mixin) in the lit-mail app. for now, we haven't had the need for types in the event handler, but I see the use case @daKmoR described.

proposals/pending-state.md
justinfagnani: Add notes about aborting and preventing defaults. I think AbortSignal from task initiators is the way to go for task cancellation, and for preventDefault() I think it should be related to the UI affordance of the task, not execution of it.

I think it's likely that a component that is itself initiating a task is doing so via some more meaningful event than a generic pending-task, and that event can have a more meaningful preventDefault().

jorenbroekema:
Hopefully this isn't too off-topic, but I read about React.Suspense and experimented with something to replicate that in vanilla webcomponent-land. Then Justin showed me this proposal and I aligned my experiment with the proposal. I figured it might be worth sharing here to show a real-life use case of where this protocol can be useful.

[suspense-element](https://github.com/jorenbroekema/suspense-element) is a wrapper element that implements this proposal and handles displaying user-projected fallback content (e.g. a spinner) or error content if any of the pending tasks promises end up throwing, so it goes one step further on this proposal by adding a handler for what to do if work is pending, completed, or completed with an error.

proposals/pending-state.md
```diff
@@ -0,0 +1,163 @@
...
+This protocol represents three of the states (Started, Completed, and Failed) with a promise-carrying DOM event.
+
+## The `pending-task` event
+
+Components with pending tasks indicate so by firing a composed, bubbling, `pending-task` Event, with a `promise` property:
+
+TypeScript interface:
+```ts
+interface PendingTaskEvent extends Event {
+  complete: Promise<void>;
```
jorenbroekema, line 47:
```suggestion
  complete: Promise<void> | Promise<void>[];
```
This would make it possible to bundle multiple pending tasks in a single event, so that we don't show the spinner when one task has completed but there are other pending tasks still pending. Then on the indicator's end it would be something like `Array.isArray(e.complete) ? await Promise.all(e.complete) : await e.complete;`

Alternatively, I suppose you could do `this.dispatchEvent(new PendingTaskEvent(Promise.all(this.pendingTaskOne, this.pendingTaskTwo)));` which is perhaps simpler 🤔 I don't have a strong preference but I thought I'd bring up the question of handling multiple pending tasks

justinfagnani: As we discussed in chat: I don't think that we want the event to vary on the type of `.complete`. If dispatchers have multiple pending tasks they can either send multiple events or call `Promise.all()` themselves.

justinfagnani:
@Westbrook now that we have the stages outlined in the readme, can we merge this straight to Draft and iterate in issues?

Westbrook:
I think that’s the benefit of the status levels. If there are any conversations worth closing before moving on or copying in some way to issues, it would be great not to lose context, however.

Westbrook (approved):
As a “Draft” proposal this looks great and opens a lot of interesting conversations as we continue to formalize the protocol.

justinfagnani:
Great. I'll merge and we can open issues for conversations.

proposals/pending-state.md
justinfagnani: I'll open an issue for this.

proposals/pending-state.md
justinfagnani: I'll open an issue for us to document guidance on events and other interfaces.

PR #5, opened 2020-10-28
Flesh out README.md for more casual visitors

Westbrook:
More a rough intro than anything else that I'd expect to be replaced later. But, the empty README left a lot to the imagination in the near term.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,14 @@
 # community-protocols
-Cross-component coordination protocols
+As the diversity of custom elements grows wider and deeper, so do the benefits of those elements being able to communicate with each other. Making this possible via reusably well-known protocols can enable components not just from X vendor to communicate with themselves, but also to interoperate with components from Y vendor and Z vendor. This project looks to define community protocols for cross-component coordination that reduce the overhead required to make that possible.
+
+Think of things like a context API, async work (Suspense-like), SSR-interop, incremental hydration, etc., many of these capabilities get locked into framework APIs that increase the barriers to interoperability. With a community defined protocol, these features and more become attainable and portable across the web components ecosystem.
+
+## Get involved
+Check out the [Issues](https://github.com/webcomponents/community-protocols/issues) and [PRs](https://github.com/webcomponents/community-protocols/pulls) currently open on this repo to join in on conversations already inflight regarding a number of exciting areas. If you have ideas on a protocol you like raised to the community open a new [issues](https://github.com/webcomponents/community-protocols/issues/new) so that authors and consumers of libraries and components from across the community that likely have (or wanted to have) implemented features in that area can join in with their use cases as well. Once the rough edges have been hammered out, submit a PR with specs of the protocol and append it to the "Protocols" list below so it can be formalized and put into use across the community.
+
+
+## Proposals
+
+| Proposal        | Author         | Status |
+|-----------------|----------------|--------|
+| TK | TK | TK     |


README.md
```diff
@@ -1,2 +1,15 @@
 # community-protocols
-Cross-component coordination protocols
+While natively providing the component model and script/style encapsulation that makes web native componentized development possible, these capabilities can also lead to elements that are worlds (possibly "frameworks) unto themselves. While there are use cases where this is _exactly_ the super power that web components intended to bring to the web, there are as many (if not more) where this can cause a headache for web component consumers looking to leverage a selection of components without needing prior knowledge of their internals or the tools that brought them into being. To this end, community protocols for cross-component coordination will expand on the default interoperability of web components all being based at the root on browser APIs by clarifying to component authors paths for communication between elements of dispirate implementation in a way that further empowers component consumers to develop more powerful applications.
```
justinfagnani, line 2: I would prefer to trim this intro down to a pretty perfunctory description of the need to have coordination between components of different origins. A sentence or two should do for now and we can build up more detailed motivation later.

Westbrook: Updated version work for ya?

README.md
```diff
@@ -1,2 +1,15 @@
 # community-protocols
-Cross-component coordination protocols
+As the diversity of custom elements grows wider and deeper, so does the benefits of those elements being able to communicate with each other via reusably well-known protocols that ensure components from X vendor can interoperate with components from Y vendor. This project looks to define community protocols for cross-component coordination that reduce the overhead required to make that possible without the side effect of locking consumers of those elements into specific patterns themselves.
+
+Think of things like a context API, async work (Suspense-like), SSR-interop, incremental hydration, etc., many of these capabilities get locked into framework APIs that increase the barriers to interoperability. With a community defined protocol these features and more could become more plug-n-play, more opt-in, and more individually susseptible to the sort of competition that drives innovation and change for the better on the web.
```
EisenbergEffect, line 4: The final sentence of this paragraph could use some rework I think. The word "more" is used a bit much. Also "plug-n-play" and "opt-in" seem to be getting at the same core idea. Perhaps that can be made more succinct. Finally, the final clause of the last sentence appears to imply that conformance is a catalyst for innovation, which I don't find much evidence for. Perhaps I've misunderstood the idea there. I think that should be reworded or perhaps dropped.

README.md
```diff
@@ -1,2 +1,15 @@
 # community-protocols
-Cross-component coordination protocols
+As the diversity of custom elements grows wider and deeper, so does the benefits of those elements being able to communicate with each other via reusably well-known protocols that ensure components from X vendor can interoperate with components from Y vendor. This project looks to define community protocols for cross-component coordination that reduce the overhead required to make that possible without the side effect of locking consumers of those elements into specific patterns themselves.
```
EisenbergEffect, line 2: This paragraph is made up of two rather long sentences. They feel a but like run-ons. Perhaps break them up into shorter sentences. There are also some grammar errors, such as "so does the benefits" instead of something like "so does the benefit" or "so do the benefits".

Westbrook: Ran this by grammarly, let me know what you think!

README.md
Westbrook: Makes sense.

justinfagnani:
I forgot about this. Merging to give any visitors a clue to what's going on :) Thanks @Westbrook !

Issue #2, opened 2020-10-28
Context API


justinfagnani:
The Lit team has been prototyping a Context-like API based on events.

The basic idea is that components that need some contextual data will fire an event to request the data. The event will carry a callback use the pass the data to the requesting component. Components that can provide the data will respond to the event and call the callback with the data. The callback can then trigger arbitrary work in the requesting component, including a re-render.

The details needed for interop are things like the event name, callback property name and signature, and any keys that are used to identify the context and or data object.

justinfagnani:
@sorvell might be able to put up a draft PR with a few more details relatively soon

brucou:
That's awesome! I have been thinking about component architecture for some time. Will be happy to comment on the PR when you have it. I should maybe try to explicit a conceptual framework and vocabulary going forward. 

~The basic ideas are that components have different communication needs:~
~1. some components are standalone, independent etc. they do not require anything from the rest of the world to achieve their task. These are the typical pure view components~
~2. some components may request (fetch data) or change (send a command to some remote device) the outside world. This, still in a standalone manner, meaning without requiring collaboration or cooperation from other components.~
~3. some components realize their task through limited collaboration with other components.~
~4. some components realize their task through extensive bi-directional collaboration with other components. In other words, the communication is complex enough to deserve to be called a protocol~

If we have two distinct and separate local entities (i.e. they are both local, but they share nothing locally) and need to coordinate to realize a task, two possibilities are event broadcasting, and message passing (we just excluded the third one which communication through shared state). The difference between the two resides in that with message passing, the recipient's address is known when sending the message, while with event handling only the event being sent is known. Those are precisely what you suggest: communication through callback is message passing, and firing events is event broadcasting.

Some issues with event broadcasting:
- the event is not guaranteed to be received.
- the event is not guaranteed to be received by the entity who should receive it.
- the event is not guaranteed to be received only by the entity who should receive it.
- the event couples sender and receiver(s) ([data or stamp coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)), arguably the weakest form of coupling). Versioning is thus worth thinking about. If an event changes shape then all consumers must adapt to the new event shape. So if you ship the event-coupled entities separately, you need to define compatibility between versions.

Some issues with message passing:
- the recipient may not be at his address (e.g., the recipient may have been disposed of)
- same coupling issue than with event broadcasting

[...]

PR #10, opened 2021-04-27
[context] Initial draft

benjamind:
This adds a formal PR for the Context API Proposal as outlined in #2

### DIFF
--- a/proposals/context.md
+++ b/proposals/context.md
@@ -4,9 +4,9 @@ An open protocol for data passing between components.
 
 Author: Benjamin Delarre
 
-Document status: Proposal
+Document status: Draft
 
-Last update: 2021-6-11
+Last update: 2021-6-25
 
 # Background
 
@@ -38,54 +38,87 @@ State management libraries often need to perform similar behaviors to the proble
 
 At a high level, the Context API is an event based protocol that components can use to retrieve data from any location in the DOM:
 
-- A component requiring some data fires a `request-context` event.
-- The event carries a `name` that denotes the data requested, and a `callback` which will receive the data.
-- Providers can attach event listeners for `request-context` events to handle them and provide the requested data.
+- A component requiring some data fires a `context-request` event.
+- The event carries a `context` value that denotes the data requested, and a `callback` which will receive the data.
+- Providers can attach event listeners for `context-request` events to handle them and provide the requested data.
 - Once a provider satisfies a request it calls `stopPropagation()` on the event.
 
 # Details
 
-## The `request-context` event
+## The `context-request` event
 
-Components which wish to receive some data from their ancestors should initiate the request by firing a composed, bubbling, `request-context` Event, with a `callback` property.
+Components which wish to receive some data from their ancestors should initiate the request by firing a composed, bubbling, `context-request` Event, with a `callback` property.
 
 Typescript interface:
 
 ```typescript
-interface RequestContextEvent extends Event {
+interface ContextEvent<T extends Context<unknown>> extends Event {
   /**
    * The name of the context that is requested
    */
-  readonly name: T;
+  readonly context: T;
   /**
-   * A boolean indicating if the context should only be provided once.
+   * A boolean indicating if the context should be provided more than once.
    */
-  readonly once: boolean;
+  readonly multiple?: boolean;
   /**
    * A callback which a provider of this named callback should invoke.
    */
-  readonly callback: RequestContextCallback<ContextTypeMap[T]>;
+  readonly callback: ContextCallback<T>;
 }
 ```
 
 A full typescript definition for this event and its associated types can be found in the [Definitions](#definitions) section at the end of this document.
 
+## The `createContext` function
+
+Implementations should provide a `createContext` function which is used to create a well known value to identify a specific Context. These well known values could be published in lightweight NPM packages to facilitate community compatibility.
+
+It is suggested the `createContext` implementation take a string identifier to facilitate debugging, and an optional initial value which consumers could use if no Context is provided. The function should return a `Context` object with the following type definition:
+
+```ts
+export type Context<T> = {
+  name: string;
+  initialValue?: T;
+};
+```
+
+It is proposed that the `community-protocols` repository will publish a package that contains a default `createContext` implementation. An example implementation is as follows:
+
+```ts
+export type UnknownContext = Context<unknown>;
+
+export type ContextType<T extends UnknownContext> = T extends Context<infer Y>
+  ? Y
+  : never;
+
+export function createContext<T>(name: string, initialValue?: T): Context<T> {
+  return {
+    name,
+    initialValue,
+  };
+}
+```
+
 ## Context Providers
 
-A context provider will satisfy a `request-context` event, passing the `callback` the requested data whenever the data changes. A provider will attach an event listener to the DOM tree to catch the event, and if it will be able to satisfy the request _MUST_ call `stopPropagation` on the event.
+A context provider will satisfy a `context-request` event, passing the `callback` the requested data whenever the data changes. A provider will attach an event listener to the DOM tree to catch the event, and if it will be able to satisfy the request _MUST_ call `stopPropagation` on the event.
 
-If the provider has data available to satisfy the request then it should immediately invoke the `callback` passing the data. If the event does NOT have a truthy `once` property, then the provider can assume that the `callback` can be invokved multiple times, and may retain a reference to the callback to invoke as the data changes. If this is the case the provider should pass the second `dispose` parameter to the callback when invoking it in order to allow the requester to disconnect itself from the providers notifications.
+If the provider has data available to satisfy the request then it should immediately invoke the `callback` passing the data. If the event has a truthy `multiple` property, then the provider can assume that the `callback` can be invokved multiple times, and may retain a reference to the callback to invoke as the data changes. If this is the case the provider should pass the second `dispose` parameter to the callback when invoking it in order to allow the requester to disconnect itself from the providers notifications.
 
 A provider does not necessarily have to be a Custom Element, but this may be a convenient mechanism.
 
 ## Usage
 
-An element which wishes to receive some context and participate in the Context API should emit an event with the `request-context` type. It is suggested that an implementation of the `RequestContextEvent` would be used something like this:
+An element which wishes to receive some context and participate in the Context API should emit an event with the `context-request` type. It is suggested that an implementation of the `ContextEvent` would be used something like this:
 
 ```js
+// get a context from somewhere (this could be in any module)
+const coolThingContext = createContext('cool-thing');
+
 this.dispatchEvent(
-    new RequestContextEvent(
-        'cool-thing', // the name of the context we want to receive
+    new ContextEvent(
+        coolThingContext, // the context we want to retrieve
         callback: (coolThing) => {
             this.myCoolThing = coolThing; // do something with value
         }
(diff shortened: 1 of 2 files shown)


benjamind:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fwebcomponents%2Fcommunity-protocols%2Fpull%2F10) for more info**.

@googlebot I signed it!

proposals/context.md
```diff
@@ -0,0 +1,265 @@
+# context-protocol
```
justinfagnani, line 1: I think we can go with title-cased names here as the name of the protocol.

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+
+# Non-Goals
+
+**Context API !== Dependency Injection Framework**
+
+The Context API does not intend to cover all cases and forms of Dependency Injection. It does not specify constructor, factory or property injection patterns. Its only goal is to formalize the pattern of sharing data across the hierarchy in the DOM, specifically avoiding 'prop drilling' type scenarios. Dependency Injection patterns could be implemented using this protocol, but this is not the goal and should remain explicitly outside the scope of Context API for simplicity.
+
+**Context API is not a state management alternative**
+
+State management libraries often need to perform similar behaviors to the problems that Context API helps to solve. An element deep in the DOM tree made need access to some state, and may need to respond to that state being changed. While state management could be built using the Context API, it is not a primary goal of Context API to solve this problem.
```
justinfagnani, line 35: Might be worth a mention that context is often useful for wiring up state management, ie how Redux's useSelector() uses useContext() to get a store.

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+
+# Details
```
justinfagnani, line 37: I think it's be nice to have a brief overview before the details, which describes at a very high level the approach of firing an event to represent a request for a context object, that the event carries a callback to provide values, and that any listener above the dispatcher in the tree can handle the event, provide an object, and stop propagation.

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+Components which wish to receive some data from their ancestors should initiate the request by firing a composed, bubbling, `request-context` Event, with a `callback` property.
+
+Typescript interface:
+
+```typescript
+interface RequestContextEvent extends Event {
+  /**
+   * The name of the context that is requested
+   */
+  readonly name: T;
```
justinfagnani, line 50: I think there was some outstanding discussion around string names vs Symbol, yeah? I don't quite remember where that left off. Can we add a todo and an issue? I see some discussion on name conflicts below, but it'd be nice to directly address whether a string key is the best approach and alternatives considered.

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+   */
+  readonly callback: RequestContextCallback<ContextTypeMap[T]>;
+}
+```
+
+A full typescript definition for this event and its associated types can be found in the [Definitions](#definitions) section at the end of this document.
+
+## Context Providers
+
+A context provider will satisfy a `request-context` event, passing the `callback` the requested data whenever the data changes. A provider will attach an event listener to the DOM tree to catch the event, and if it will be able to satisfy the request will call `stopPropagation` on the event.
```
justinfagnani, line 66: Should we tighten the language and say that in order to be compliant with the protocol, a listener _must_ call `stopPropagation()` if it provides a value now, or will in the future? I can't think of where it would make sense to have multiple providers.

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+
+If the provider has data available to satisfy the request then it should immediately invoke the `callback` passing the data. If the event does NOT have a truthy `once` property, then the provider can assume that the `callback` can be invokved multiple times, and may retain a reference to the callback to invoke as the data changes. If this is the case the provider should pass the second `dispose` parameter to the callback when invoking it in order to allow the requester to disconnect itself from the providers notifications.
```
justinfagnani, line 68:
```suggestion
If the provider has data available to satisfy the request then it should immediately invoke the `callback` passing the data. If the event does NOT have a truthy `once` property, then the provider can assume that the `callback` can be invokved multiple times, and may retain a reference to the callback to invoke as the data changes. If this is the case the provider must pass the second `dispose` parameter to the callback when invoking it in order to allow the requester to disconnect itself from the providers notifications.
```

Add "must"? I think the key requirement is that if a provider may call the callback multiple times, it must provide a disconnect function, so that the converse is true: a defined disconnect function is an indicator of whether or not the callback can be called multiple times.

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+
+A provider does not necessarily have to be a Custom Element, but this may be a convenient mechanism.
+
+## Usage
+
+An element which wishes to receive some context and participate in the Context API should emit an event with the context-request type. This library includes an implementation of the ContextEvent class which implements the above interface and can be used in the following way:
```
justinfagnani, line 74:
```suggestion
An element which wishes to receive some context and participate in the Context API should emit an event with the context-request type. This library includes an implementation of the RequestContextEvent class which implements the above interface and can be used in the following way:
```

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+        }
+    )
+);
+```
+
+If a provider listening for this event can provide the requested context it will invoke the callback passed in the payload of the event. The element can then do whatever it wishes with this value.
+
+It may also be the case that a provider can retain a reference to this callback, and can then invoke the callback multiple times. In this case providers should pass a dispose function as a second argument to the callback to allow consumers to inform the provider that it should no longer update the element, and should dispose of the callback.
+
+As a convenience, and a hint for providers, an element may also provide a once boolean on the event detail to indicate that it is not interested in receiving updates to the value. If this behavior is essential to the correct operation of the consumer, then they should be implemented defensively as there is no guarantee that providers will honor this agreement. An example is provided below:
```
justinfagnani, line 91: It's more than just a hint, isn't it? If a consumer only wants a value once, then the provider doesn't have to keep the callback in memory or pass in a disconnect callback.

Along those lines, should `once` should default to true (or the option be `multiple: false`) in order to default to the lower-memory, less chance of a memory leak case?

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+    this.logger = undefined;
+  }
+}
+```
+
+# Open Questions / Previous considerations
+
+## Could we use Promises instead of callbacks?
+
+Many have commented that Promises could be used instead of callback functions. The major drawback of promises is that they do not allow multiple-resolution. Therefore we would not be able to handle cases where a requested value changes over time. This is capability that we see in the React Context API and believe is valuable for a variety of use cases.
```
justinfagnani, line 145: Promises can also not be resolved synchronously

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+
+While we could restrict this API to only support a single resolution of a requested value, and then use observable mechanisms on that value to achieve data update behaviors, it is believe this will complicate simple use-cases unnecessarily.
+
+## What about a `createContext` style API?
+
+In React Context API they provide a function `createContext` which creates a context object that uniquely identifies the context to consumers and providers. This has some advantages in that it means you have a unique object that identifies a context. However the downside is that you need to have an agreed upon module that exports this context value in order to share the context.
+
+For web components this creates a centralization issue. If all components want is a logger context, where would the logger context object be defined? Who owns that module? Will all consumers agree to depend upon it?
```
justinfagnani, line 153: If all requestors of a logger context are expecting a particular API, isn't it better if they indicate that by using a shared module? What happens if multiple consumers depend on incompatible logger interfaces keyed by the same context name?

proposals/context.md
```diff
@@ -0,0 +1,265 @@
...
+  // test if the provider is the type we want or the value it provides is right
+});
+const dispose = provider.provide(this, (logger) => {
+  this.logger = logger;
+});
+// later...
+dispose(); // don't need updates anymore
+```
+
+These alternatives do provide more capability, but its an open question as to whether or not this complexity is warranted or desired. It also opens up a larger question about what would the candidate value be? Would it have to be an object of the requested type, could it be some other protocol to determine uniformity between the requested data and the actual data? This begins to seem more complex than we really need here for unnecessary type safety overhead. It is suggested if consumers want type safety then they should use Typescript to achieve this.
```
justinfagnani, line 205: Have we seen a real use-case here yet or an analogy in a framework? I agree that implementing a runtime-typecheck callback seems overly complex. If really needed, this would be an argument for more guaranteed unique context keys.

proposals/context.md
benjamind: Yes we haven't really dug into this too far. I've tried to theorize about how things would play out if we used object-values (Symbol, or unique value returned from a 'create' function), but these have all seemed to require some agreed upon central repository of context object keys. This seems somewhat restrictive?

justinfagnani: I think restrictive is the point. Unless you have some agreement to provide a compatible object, context is going to be too fragile. Importing the same Symbol, which can host docs on the type of object it represents, is a basis for agreement.

I don't think there would need to be any central repository either. All you need is access to the same module. That can happen anywhere, even in a one-liner npm package. That package will serve as the docs for what the context object is supposed to be. If that seems heavyweight, I'd say that even with string keys, if you have cross package collaboration, _something_ needs to describe what that object should be, and that thing can host the symbol.

proposals/context.md
benjamind: I agree its definitely more strict to work that way.

Can we hope for a consistent implementation though across the WC community for common contexts? I think if we could have the community-protocols repository produce some libraries of common contexts and their associated API shapes this might work well perhaps?

proposals/context.md
benjamind: Changed to `Context Protocol` in bccaeac

proposals/context.md
benjamind: Agreed, added some more detail here bccaeac

proposals/context.md
benjamind: Gave a bullet point list outline in bccaeac

proposals/context.md
benjamind: That seems reasonable to me.

proposals/context.md
benjamind: Done in bccaeac

proposals/context.md
benjamind: Revised this text in bccaeac

proposals/context.md
benjamind: Expanded on this a little in bccaeac

proposals/context.md
benjamind: After some offline discussion an outline of how this might look is as follows, new definitions

```typescript
export type Context<T> = {
    key: string | symbol;
    _type?: T; // flavored brand for the type system
};

export type AnyContext = Context<unknown>;

export function createContext<T>(key: string | symbol): Context<T> {
    return {
        key,
    };
}
```

These would then be used something like this to define a new context and export it:

```ts
// define an interface for the context value
export interface Logger {
    log: (msg: string) => void;
};
// create a context object
export const loggerContext = createContext<Logger>('logger');
```

Components, applications and libraries which produce and consume contexts should publish modules containing them, preferably small atomic modules with just interfaces and the context object. Perhaps common contexts could be agreed upon and interfaces published by groups like the community-protocols group.

The context could then be requested like so:

```ts
import { loggerContext, Logger } from '@community-protocols/contexts/logger'; // a common package

// later a component could get the context via event
this.dispatchEvent(new RequestContextEvent(loggerContext, (logger) => { this.logger = logger; }));
```

This has one major advantage over the string based keys in that it completely avoid namespace collisions. It also simplifies usage of the type system in Typescript avoiding the need for interface map merging.

I remain a little concerned that this does open the possibility for more and more functionality to be added to the context objects returned by `createContext` and that this might make compatibility harder over time as libraries adopt divergent implementations. But I do like the increased ease of type safety, and the name collision avoidance.

Westbrook (commented):
@justinfagnani if @benjamind cleans up the conflict in README, is this good to move forward for further review/work in issues/breakout sessions?

justinfagnani:
I think @benjamind had a pending change to make around the context key. It's very close either way.

proposals/context.md
benjamind: Agreed on the shift to `multiple` over `once`, have made this change in lit-context implementation, will update the proposal doc this afternoon.

benjamind:
Yes I will revise the proposal to match the implementation we have in lit context around the contextKey today and update this PR to be ready to merge.

proposals/context.md
benjamind: I agree there isn't a strong use case for these yet, and with the tighter binding to use context value objects rather than strings I think this helps avoid the need for validation?

proposals/context.md
benjamind: Have updated the proposal to match the above in  66ec2ad

benjamind:
Have updated the proposal, I believe this is ready to merge as Draft and then further revise via issue tickets?

justinfagnani:
Merging and taking up issues sgtm!

PR #11, opened 2021-06-17
Initial status outline

Westbrook:
Outline the various status level at which a protocol might currently be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,23 @@ Check out the [Issues](https://github.com/webcomponents/community-protocols/issu
 | Proposal        | Author         | Status |
 |-----------------|----------------|--------|
 | TK | TK | TK     |
+
+## Status
+
+Community Protocols will go through a number of phases during which they can gather further community insight and approval before becoming "Accepted". These phases are as follows:
+
+- *Proposal*
+  "Proposal" status applies to just about anything that community members are interested in putting some thought into. While an issue submitted to this repo can help generate initial ideas on the protocol or space of interest and clarify the information needed to kick off a more fruitful conversation, a PR will serve to make known that you are interested in support to drive the protocol in question forward. Having an initial explainer included in this PR, while adding the protocol to the "Proposals" table shown above, will prepare the community to both communicate about and contribute to the development of the protocol asynchronously.
+
+- *Draft*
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,3 +12,23 @@ Check out the [Issues](https://github.com/webcomponents/community-protocols/issu
...
 | TK | TK | TK     |
+
+## Status
+
+Community Protocols will go through a number of phases during which they can gather further community insight and approval before becoming "Accepted". These phases are as follows:
+
+- *Proposal*
+  "Proposal" status applies to just about anything that community members are interested in putting some thought into. While an issue submitted to this repo can help generate initial ideas on the protocol or space of interest and clarify the information needed to kick off a more fruitful conversation, a PR will serve to make known that you are interested in support to drive the protocol in question forward. Having an initial explainer included in this PR, while adding the protocol to the "Proposals" table shown above, will prepare the community to both communicate about and contribute to the development of the protocol asynchronously.
+
+- *Candidate*
```
justinfagnani, line 23: I thought we had said that candidate was after draft and the last call for feedback before becoming accepted, somewhat like TC39: https://tc39.es/process-document/

justinfagnani (changes requested):

README.md
Westbrook: I may have my notes wrong here. I actually have two swap marks on that as I remember going back and forth on it last minute. Does anyone else have a preference? Happy to swap it if that's the correct path!

justinfagnani: I think aligning with TC39 will be easiest, since it's pretty well-known.

Westbrook: Swapped.

Westbrook:
Question: 

- Did we want to include reference to "Rejected" (maybe by a slightly softer word) proposals for the record, or could these just _stay_ in whatever status the get blocked in?

justinfagnani:
I think we should have such a state. I can image a lot of potential protocols ideas that are useful, but maybe too niche to be included here, for instance. I think we should try to define the scope of the repo a bit better, along with some other criteria (being able to be implemented independently), then maybe add "Out of Scope", "Complexity", etc reasons for rejection

justinfagnani (approved):

justinfagnani:
Merging this now. We can add rejected states and guidance on where to put proposals in the source tree later.

PR #15, opened 2021-06-24
[defer-hydration] Initial draft

justinfagnani:


proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+
+Status: Draft
+
+Last update: 2021-06-24
+
+# Background
+
+In server-side rendered (SSR) applications, the process of a component running code to re-associate its template with the server-rendered DOM is called "hydration". The defer-hydration protocol is design to allow controler over hydration to solve two related problems:
+
+1. Interoperable incremental hydration across web components. Componentents should not automatically hydrate upon being defined, but wait for a signal from their parent or other coordinator.
```
matthewp, line 15: Can you explain why this is (why they shouldn't automatically hydrate)? Is there an example where doing so would create unwanted results?

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+2. Hydration ordering independent of definition order. Because components usually depend on data from their parent, and the parent won't usually set data on a child until it's hydrated, we need hydration to occur in a top-down order.
```
matthewp, line 16: I guess this is maybe a partial answer for (1)? Should these be separate points, if so?

proposals/defer-hydration.md
Westbrook: I could see this fitting into a "when work gets done" topic of conversation by allowing islands to be linked generally by component name, while still being able to disambiguate on which of those islands are _needed_ at any one time. Much like Astrobuild leverages the `:visible` modifier to control when the code for a component is loaded, this could allow other instances of the component on the same page the ability upgrade with the initially visible instance, and then wait to hydrate until they, too, are in view.

I think this sort of laziness deserves to be revisited more thoroughly a the browser spec level. This seems roughly in line with the a proposal around lazy definitions: https://github.com/WICG/webcomponents/issues/782 and could arguably fit into a larger conversation around the browser allowing for lazy upgrades, which would be somewhat similar to what this protocol looks to support.

Regardless, if you're SSR'ing the entirety of an element that _may_ not come into view, the ability to control when it reconciles that delivered SSR render with the client-side render sounds like a good idea.

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+        <x-figure defer-hydration>...</x-figure>
+      </template>
+    </x-article>
+  </body>
+<html>
+```
+
+## Use case 2: On-demand hydration
+
+Hydration can be deferred until some data or user-driven signal, such as interacting with an element. In this case server-rendering is configured to add `defer-hydration` to all elements so that nothing will automatically hydrate.
```
Westbrook, line 60: `visibility` is such a hot/important topic for upgrade/hydration that we should probably include it in line with `such as interacting with an element`.

Westbrook (commented):
@justinfagnani this looks like a great start.

What are you thoughts on this possibly being paired with a conversation at the spec level for revisiting lazy definitions, lazy upgrading, etc.? 

Also, @liamdebeasi I wonder how you might see something like this aligning with the lazy loading strategy that Stencil uses. Feel free to ring in any of your teammates that would be better fit to the conversation, if you'd like.

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+
+Author: Justin Fagnani
+
+Status: Draft
+
+Last update: 2021-06-24
+
+# Background
+
+In server-side rendered (SSR) applications, the process of a component running code to re-associate its template with the server-rendered DOM is called "hydration". The defer-hydration protocol is design to allow controler over hydration to solve two related problems:
```
thescientist13, line 13: A couple minor grammatical point(s)

change from
> _The defer-hydration protocol is design to allow controler over hydration to solve two related problems:_

to
> _The defer-hydration protocol is ~~design~~ **designed** to allow ~~controler~~ **control** over hydration to solve two related problems:_

proposals/defer-hydration.md
matthewp: The more I think about it, I don't think I agree with this point (2), cc @Westbrook.  I would expect that in most cases an element's initial data has been serialized as attributes and doesn't need the parent to pass that information. The exception would be data that isn't serialized, like complex objects, but I don't think that's really the norm.

Westbrook: The https://github.com/w3c/webcomponents-cg has a breakout session to discuss Community Protocols, that could certainly include this, at the beginning of next month. Would you be interested in joining the call? Would be great to get your thoughts on this, and many other of the topics we're discussing right now "in person"!

matthewp: Yeah I could probably attend that.

Just for clarity, I'm not against this proposal, just trying to understand the scenarios in which it is needed. The text could help by clarifying that a bit more.

My thinking is that a typical usage would not need this because:

1. Static imports take care of making sure children are hydrated first.
2. Even if hydration happens out-of-order state should be serialized in attributes.

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+
+1. Interoperable incremental hydration across web components. Componentents should not automatically hydrate upon being defined, but wait for a signal from their parent or other coordinator.
+2. Hydration ordering independent of definition order. Because components usually depend on data from their parent, and the parent won't usually set data on a child until it's hydrated, we need hydration to occur in a top-down order.
+
+`defer-hydration` enables us to decouple loading the code for web component definitions from starting the work of hydration, and enables top-down ordering and sophisticated coordination of hydration, including triggering hydration only on interaction or data changes for specific components.
+
+# Overview
+
+The Defer Hydration Protocol specifies an attribute named `defer-hydration` that is placed on elements, usually during server rendering, to tell them not to hydrate when they are upgraded. Removing this attribute is a signal that the element should hydrate. Elements can observe the attribute being removed via `observedAttribute` and `attributeChangedCallback`.
```
splitinfinities, line 22: Curious on folks perspective around `defer-hydration` as compared to something akin to `loading="lazy"`

If the attribute was, say, `hydration`, with options like: 
 - `defer`: Hydrates based on internal behavior
 - `lazy`: Hydrates based on when in viewport
 - `eager`: Default behavior
 
Should the attribute control both the intended behavior and the current state? Some tools use `:visible`, Stencil uses the `hydrated` class for hydrated state. 

Decoupling this and setting a convention around the intended behavior and leaving how state is communicated up to the tools might be helpful.

daKmoR:
an additional behavior that might be nice to have is `hydration="never"`. For example if you have purely visual web components.

matthewp:
I saw that 11ty had implemented this proposal so I wanted to reiterate my questions. What is missing from this proposal is a concrete use-case. What is the use-case where a component should delay its hydration because a parent told it to? What is the parent needing to do to the component before it hydrates?

A good use-case would be something like a special type of component that is tightly coupled to its parent. Point (2) specifies data dependencies, but that is not what I would call a specific use-case.

JosefJezek:
https://www.zachleat.com/web/defer-hydration/

justinfagnani:
The biggest motivation for this proposal is to ensure top-down initialization of SSR'ed elements.

In client-side code, shadow children are necessarily created after their parents. So they will have received properties from their parents and can assume that they can fire events that will be handled by their parents.

With SSR'ed HTML and shadow roots elements are already created and will initialize in _definition_ order. This can cause three major problems:
1. A polynomial amount of initialization work. If a child initializes before its parent, it will use default values for properties and render. If a parent then initializes, the parent will use default value for _its_ properties, then write properties into the child, causing the child to render again. If a grandparent then initializes it will do the same, causing the the parent and child to render again.
  - 1a: Aside from extra work, mismatches between SSR'ed state and client state are sometimes very bad for hydration, like if a certain hydration pass expects to be given the same data as used in SSR as an optimization.
2. Event listening and dispatch can happen out-of-order causing missed events. defer-hydration lets components wait until their parents are ready and have had a chance to add listeners before dispatching events

In addition to those problems that defer-hydration solves, it also creates an interoperable signal for islands-type architectures to use to trigger components to hydrate by removing the attribute on visibility, interaction, etc,.

matthewp:
Thanks! I think I see 2 use-cases here. 1 is a tight coupling between components that are not serialized in such a way that they could be hydrated independently. The second is the island-control use-case. I'd like to skip the island idea for now and focus on the first idea because it's the one I understand the least.

Let's put this to real-code. Let's say you have this:

```html
<parent-1>
  <template shadowroot="open">
    <child-1></child-1>
  </template>
</parent-1>
```

Correct me if I'm wrong, but I think you are saying that `<child-1>` will render first with its defaults and then again when `<parent-1>` hydrates with possibly different properties being passed down to `<child-1>`. And the same would occur again if there were grandparents, etc.

If I'm understanding this correctly, then this can be fixed by serializing `<child-1>` with the properties (attributes) it needs:

```html
<parent-1>
  <template shadowroot="open">
    <child-1 first-name="Matthew"></child-1>
  </template>
</parent-1>
```

So my next question is why _wouldn't_ you do it this way? I think I'm missing something, and there's a reason why you can't just serialize.

kevinpschaaf:
> So my next question is why wouldn't you do it this way? 

This is indeed a partial solution, but it only works will easily serializable primitive data for e.g. leaf widgets, and breaks down with more complex application data:

- Not all data passed from parent to child is (easily) serializable, in the case of JS property-valued items (structured data, classes.)
- Even if you came up with a serialization method for all data (for now, just assume JSON or similar), serializing real application data that is prop-drilled down the tree would be wildly inefficient (grandparent gets `{a: {b: {c: 5}}}` serialized to it, parent gets `{b: {c: 5}}`, child gets `{c: 5}` etc... with polynomial fanout down the tree)
- Deserializing e.g. JSON necessarily means creating new object instances; thus two children in the tree having been passed the same object will end up with separate object instances, which may break assumptions in the application.

This protocol is attempting to help provide a robust SSR capability for Web Components in the general case. If a given SSR implementation knew all of the data for a given WC instance was serializable to attributes, it could opt to let the instance eagerly hydrate and not use `defer-hydration`.

Thus, this protocol is addressing cases where that's not possible/practical.

matthewp:
Thanks! So the use-case is passing complex types.

Are there examples of generic components that would need complex types? Are there examples of native elements that need complex types?

My current thinking is that this use-case is for tightly coupled "application logic" components. These types of components are contained to within a single codebase. Perhaps cross-team within a company. But not a generic component that you might distribute to just anyone.

With that being the case I don't think a community standard is necessary here. If `<child-1>` requires a complex property to be passed to it, it can define its own API for it to wait.

----

After thinking about it more, my general take is that delaying hydration is _bad_ and should be avoided when possible. Doing something like this within your own application because you value the developer convenience of passing complex objects is completely fine, but it's not something we should endorse.

dgp1130:
As I've been playing around with this proposal, one thing which has bothered me is the fact that you need to put `defer-hydration` fully down the render tree for every custom element. This can be tricky to manage and you end up with:

```html
<my-outer-component defer-hydration>
  <template shadowroot="open">
    <!-- I *MUST* put `defer-hydration` here too, or else it will hydrate immediately. -->
    <my-inner-component defer-hydration></my-inner-component>
  </template>
</my-outer-component>
```

Since `defer-hydration` is not the default, it means you have to write it for every element. It is very easy to forget to add this to an element, and it's difficult to validate. Particularly for the island use case, you'd want everything _under_ the island to be deferred, and then when that island is hydrated, everything underneath it should be hydrated as well.

I couldn't think of a good way to do this, but I recently found out about [custom state pseudo-classes](https://css-tricks.com/custom-state-pseudo-classes-in-chrome/) which _almost_ implement this. The key thing about pseudo-classes is that they inherit down the tree just like CSS does. This means that a component at the bottom of the render tree can detect what its parent's state is. If we define it as custom state like `:--defer-hydration`, this means we can do:

```html
<my-outer-component defer-hydration>
  <template shadowroot="open">
    <!-- No `defer-hydration` attribute here! -->
    <my-inner-component></my-inner-component>
  </template>
</my-outer-component>
```

```typescript
class MyOuterComponent extends HTMLElement {
  constructor() {
    super();

    // Set the custom state based on the `defer-hydration` attribute.
    const internals = this.attachInternals();
    if (this.hasAttribute('defer-hydration')) internals.states.add('--defer-hydration');
  }
}
customElements.define('my-outer-component', MyOuterComponent);

class MyInnerComponent extends HTMLElement {
  constructor() {
    super();

    // Only hydrate if the custom state is not set.
    const defer = this.matches(':host-context(:--defer-hydration)');
    if (!defer) this.hydrate();
  }

  // ...
}
customElements.define('my-inner-component', MyInnerComponent);
```

And the inner component is able to defer based on the parent's `:--defer-hydration` state! I'd love for this to work, but there are a couple issues today:
1. Custom state pseudo classes are only supported in Chrome and not a fully supported standard.
2. `this.matches(':host-context(...)')` always returns `false` AFAICT. This does work if you use it in a real stylesheet to style something.
3. Currently the only API I'm aware of to _set_ state is to use `this.attachInternals().states.add('--defer-hydration')`. That requires the top-level element to execute and upgrade _first_ in order to set that state based on its attribute _before_ any descendent nodes upgrade, which can be difficult to enforce.
    * A way to set this state declaratively in HTML is probably required. Something like `<my-outer-component :--defer-hydration />` would be needed to enable the state without hydrating `my-outer-component`.
4. Currently we need `attributeChangedCallback()` to synchronously detect a change to `defer-hydration` and hydrate. I'm not aware of such a hook for detecting CSS. I believe you can use `MutationObserver` to detect some CSS changes, but it is async and likely not sufficient. Some kind of `customStateChangedCallback()` would likely be needed.
    * To add to this complexity, ordering of the `customStateChangedCallback()` is important. It would probably need to flow top-down so each element can choose whether to hydrate its children (do nothing) or keep deferring them (set `--defer-hydration` until some other user interaction).
5. Custom states can only be set via `ElementInternals` which is designed to be private to the custom element, meaning you can't imperatively remove `:--defer-hydration` from another element like you can with the `defer-hydration` attribute as defined already.
    * Maybe this could be solved via a community protocol, but I'm not 100% on what that would look like.

I'm not sure if this is really a good direction for `defer-hydration`, but I thought it was an interesting approach to not having to manually repeat `defer-hydration` all the way down the tree. This might be a bit of a hack of the custom states use case and maybe at that point we should just have a proper web spec for deferring hydration. Food for thought.

proposals/defer-hydration.md
matthewp: Sorry for just now getting back to this, but why would you want to delay rendering a component? The reason for those modifiers are to:

1. Deprioritize loading of the code so that more important assets can be loaded over network.
2. Deprioritize executing the component (and its dependencies) JS so that more important assets can use the CPU instead.

The reason is not to delay *rendering*. Rendering should be relatively cheap, especially if the component was prerendered, there's no work to do but set up the event listeners.

It is, of course, possible that some components are going to be more expensive than others, but that's reason for those particular components to have their own API to delay rendering, not to make it a standard that is used arbitrarily.

The goal is to make the page as interactive as possible. Delaying rendering without good reason should be avoided.

justinfagnani: Again, the main reason is to enable top-down initialization, even for spontaneous first renders. We need our components to initialize in the same order whether CSR'ed or SSR'ed, independent of definition order.

matthewp: That's a different use case (being discussed below). The island orchestration use case is another, it's why Eleventy/webc has adopted this for example.

kevinpschaaf:
> But not a generic component that you might distribute to just anyone.

@matthewp I'd challenge you to imagine a future where `<data-tree>` and `<awesome-spreadsheet>` and `<code-editor>`  _can_ be published as reusable components, that can be dropped into any application, and enjoy all the same benefits of server-rendering that buttons will. In these cases, an API with complex objects is not simply a "developer convenience," it's a practicality of making components that require configuration that exceed the limits of primitive data types (which itself is a completely arbitrary constraint).

Another angle here is that, even with "application-specific" components, there is immense value in the flexibility that web components provide in the ability to incrementally migrate a given component or codebase from one WC base class to another; protocols like this that define any extra interface points above the standard spec'ed interface ensure that's also possible.

If the extent of your vision for web components stops at `<my-button label="Simple">`, like I said you won't be a customer of this protocol, and that's _fine_. But it needn't stop defining a protocol for those who do want to make more complex interoperable reusable components.

kevinpschaaf:
@dgp1130 That would be a pretty clever approach, but unfortunately besides the defacto limitations you mentioned with `matches()` and detecting changes, `:host-context` is also a somewhat disputed part of the V1 web components spec (see https://github.com/w3c/csswg-drafts/issues/1914) and does not have good cross-browser support. Also, in our real-world usage of `defer-hydration` so far, it's important to be able to incrementally remove `defer-hydration` element-by-element down the tree rather than once and have it inherit down, since some great-great-grandparent can't reliably know when it's safe to assume hydration down the tree is complete in order to remove the state.

matthewp:
@kevinpschaaf Thanks for some more specific use-cases. I _can_ imagine these types of components as I'm the author of [one myself](https://github.com/matthewp/comic-reader). Mine does not support server rendering at the moment. If it did I can imagine I would infer if I needed to render defaults.

You can do this easily with declarative shadow DOM by checking if `this.shadowRoot` exists. If it does then that means it was server rendered and you can skip rendering its defaults.

> If the extent of your vision for web components stops at `<my-button label="Simple">`, like I said you won't be a customer of this protocol, and that's fine. But it needn't stop defining a protocol for those who do want to make more complex interoperable reusable components.

This protocol is recommending this approach for all components. From the proposal:

"1. Interoperable incremental hydration across web components. Components should not automatically hydrate upon being defined, but wait for a signal from their parent or other coordinator."

This is the source of our disagreements. Components absolutely should hydrate upon being by defined. Only in very special circumstances should they not do so.

The fact that custom elements can hydrate independent of a tree is the biggest benefit of web components. Proposing that they not do so, and instead way for some root element to first hydrate makes them more like a React tree. We should not make exceptional scenarios the default.

Keep in mind that Eleventy/webc has implemented this proposal, so they are conforming to what is in my opinion a bad default, right now.

matthewp:
My biggest issue with this proposal is that it is a blanket recommendation that all components defer to the wishes of a parent component on when to hydrate. A element should always decide for itself when to hydrate.

This is a case-by-case basis. Some native elements have similar constraints, but they always define an API in which they are in charge and the user only provides a hint. For example `` as defined on MDN:

> Defers loading the image until it reaches a calculated distance from the viewport, as defined by the browser. 

Note the important part **as defined by the browser**. That is, the browser defines when to take action, it doesn't wait for the application developer to tell it that it's time.

kevinpschaaf:
> You can do this easily with declarative shadow DOM by checking if this.shadowRoot exists. If it does then that means it was server rendered and you can skip rendering its defaults.

I don't follow. `this.shadowRoot` existing only tells you that it was server rendered. It does not tell you when and if the host is done configuring the element with any properties that may be coming. The existence of `defer-hydration` is an explicit signal from the host that it _will_ provide data, and its removal is an explicit signal that it has _completed_ providing data.

> This protocol is recommending this approach for all components.... This is the source of our disagreements. 

> My biggest issue with this proposal is that it is a blanket recommendation that all components defer to the wishes of a parent component on when to hydrate. 

I see. In general it should be possible for `defer-hydration` to be an opt-in protocol, but I think I agree that there's a possible ambiguity in whether a rendering context knows whether it should set `defer-hydration` on an element it is rendering, given that the element may not implement `defer-hydration` AND still be configurable purely from attributes. Perhaps this argues for only setting `defer-hydration` when a rendering context detects it is setting a non-serializable JS value. Or maybe it should be a thing that's set via an `ElementRenderer` interface for a given WC class (which is specific to the element being rendered) rather than from the rendering context... need to think more about whether that works.

> A element should always decide for itself when to hydrate.

Sorry, but I don't think this follows. If we can agree that an element shouldn't hydrate until it is fully configured (for performance reasons, amongst others discussed), this is never something a component can decide for itself. A component cannot know when or if a host will or has configured it, given it's just as valid for a host to do nothing to an element as to set a bunch of elements/properties. In the general case, this is only knowable by the host.

dgp1130:
> @dgp1130 That would be a pretty clever approach, but unfortunately besides the defacto limitations you mentioned with matches() and detecting changes, :host-context is also a somewhat disputed part of the V1 web components spec (see https://github.com/w3c/csswg-drafts/issues/1914) and does not have good cross-browser support.

@kevinpschaaf, I wasn't aware of implementation issues with `:host-context()`. I don't think this approach specifically relies on that though, you can use any value which inherits down the render tree such as a CSS variable instead, though setting it in HTML does get a bit weird:

```html
<my-component style="--defer-hydration: true">
  <my-child-component></my-child-component>
</my-component>
```

```typescript
class MyChildComponent extends HTMLElement {
  constructor() {
    super();

    const defer = getComputedStyle(this).getPropertyValue('--defer-hydration')!.trim() === 'true';
    if (!defer) this.hydrate();
  }
}

customElements.define('my-child-component', MyChildComponent);
```

Definitely a hack and doesn't help with listening for changes to the `--defer-hydration` variable, but it at least doesn't require `:host-context()`. Maybe `cssVariableChangedCallback()` is more amenable to browsers than a synchronous `Element.prototype.onMatchChange()`? I dunno, probably still a dead end.

> Also, in our real-world usage of defer-hydration so far, it's important to be able to incrementally remove defer-hydration element-by-element down the tree rather than once and have it inherit down, since some great-great-grandparent can't reliably know when it's safe to assume hydration down the tree is complete in order to remove the state.

I think you could still support this use case by actually writing `defer-hydration` anywhere you wanted more fine-grained control. Your tree could look like:

```html
<my-parent-component defer-hydration>
  <!-- Loads as soon as parent's `defer-hydration` is removed. -->
  <my-component></my-component>

  <!-- Still deferred when parent's defer-hydration is removed.
         `my-parent-component` can manually remove this `defer-hydration` when ready. -->
  <my-lazy-component defer-hydration></my-lazy-component>
</my-parent-component>
```

This does require knowing which components will be hydrated lazily during render time, but I guess we're already assuming that for `defer-hydration` in general, so I don't think that's too much of an ask.

Alternatively if hydration happens top down, when `my-parent-component` hydrates it could add `defer-hydration` to `my-lazy-component` to prevent it from hydrating. But that would mean that `my-lazy-component` has its `defer-hydration` state briefly cleared (when `my-parent-component`'s `defer-hydration` attribute is removed), but then later re-set (by `my-parent-component` on hydration), and this invalid state needs to avoid triggering `my-lazy-component`'s hydration. That feels way too nuanced to me and I think it's probably better to require `defer-hydration` to be explicitly authored for any subtree which needs to be deferred longer than its parent, which I think is a reasonable requirement.

kevinpschaaf:
> Definitely a hack and doesn't help with listening for changes to the --defer-hydration variable 

@dgp1130 Yeah, something like a StyleObserver would be needed to do this robustly, which doesn't exist yet.

> Alternatively if hydration happens top down, when my-parent-component hydrates it could add defer-hydration to my-lazy-component to prevent it from hydrating.

An issue here is that `defer-hydration` must be present prior to upgrade to have an effect. In a typical `a` renders `b` renders `c` tree, `a` imports `b`, and `b` imports `c`, meaning _registration_ order goes `c` then `b` then `a` due to side-effect registration happening when dependent modules load, and this goes bottom-up the module graph. Hence `a` won't have a chance to set `defer-hydration` on `b` before `b` upgrades. 

Thus in the general case, I think it's always going to be safest to apply `defer-hydration` on the server when needed.

dgp1130:
> > Alternatively if hydration happens top down, when my-parent-component hydrates it could add defer-hydration to my-lazy-component to prevent it from hydrating.
> 
> An issue here is that `defer-hydration` must be present prior to upgrade to have an effect. In a typical `a` renders `b` renders `c` tree, `a` imports `b`, and `b` imports `c`, meaning _registration_ order goes `c` then `b` then `a` due to side-effect registration happening when dependent modules load, and this goes bottom-up the module graph. Hence `a` won't have a chance to set `defer-hydration` on `b` before `b` upgrades.
> 
> Thus in the general case, I think it's always going to be safest to apply `defer-hydration` on the server when needed.

Yes, I've talked before about how import order naturally causes bottom-up hydration and this can make certain tasks more difficult. However, I think `defer-hydration` is the solution to that problem by allowing top-down hydration when it is beneficial to do so. In your example with my approach of having `defer-hydration` apply transitively down the entire render tree, then putting `defer-hydration` on `a` will prevent hydration on `b` and `c`, regardless of import ordering. Even if `c` upgrades first, it would still see that a transitive parent (`a`) has `--defer-hydration` set and skip its hydration until that property is removed.

This does imply that `a` must be deferred in order to defer any of its children imperatively, but I don't think that's strictly true, given that you don't have to define `defer-hydration` on a custom element specifically, you could put it on any element.

```html
<my-a>
  <template shadowroot="open">
    <div defer-hydration>
      <my-b>
        <template shadowroot="open">
          <my-c></my-c>
        </template>
      </my-b>
    </div>

    <div><!--  Other stuff in my-a --></div>
  </template>
</my-a>
```

In this example, `my-a` will hydrate eagerly but `my-b` and `my-c` are deferred, even if they upgrade before `my-a`. Once `my-a` removes the `defer-hydration` attribute from the `div` tag, only _then_ will `my-b` and `my-c` hydrate (possibly in a top-down fashion if we can get listening to pseudo selector / CSS variable / arbitrary style change to apply top-down).

I don't mean to distract too much from the overall `defer-hydration` proposal, I'm happy to make a separate issue to discuss whether it should apply transitively or not. I just haven't seen a convincing argument against it other than the fact that listening to hydration changes isn't really possible today (which feels solvable to me in the browser spec, but I'm not an expert in that aspect).

justinfagnani:
It's trivial for an element to observe an attribute on itself, less so on ancestors, and even less so for any ancestor in the composed tree.

The nice thing about defer-hydration is that for elements to implement the protocol ah they have to do is observe `defer-hydration` and hydrate when it's removed. They don't have to walk up DOM trees or use mutation observed.

matthewp:
> Sorry, but I don't think this follows. If we can agree that an element shouldn't hydrate until it is fully configured (for performance reasons, amongst others discussed), this is never something a component can decide for itself. A component cannot know when or if a host will or has configured it, given it's just as valid for a host to do nothing to an element as to set a bunch of elements/properties. In the general case, this is only knowable by the host.

We don't agree on this at all. An element does know when it is "fully configured" because it is the one that gets to define what "fully configured" means. If an element needs some sort of hint from its consumer then it should create a specific API just for that element, the same way that `` does through ``.

I understand that you want a generic solution for your framework so that you can take care of this problem for your users. That's fine to do within your framework. The tradeoff is that some components (many, most?) will hydrate more slowly than they need to. If you consider that an acceptable tradeoff that's within your right.

But we should not encourage a standard/recommendation that all components hydrate more slowly.

The most pressing question for all of these community standards is, imo, "would native elements adopt this"? I think the answer here is a definitive **no**.

justinfagnani:
@matthewp What do you mean by "hydrate more slowly than they need to"?

zachleat:
> Keep in mind that Eleventy/webc has implemented this proposal, so they are conforming to what is in my opinion a bad default, right now.

(As the author of Eleventy/WebC) I guess unsurprisingly I don’t agree—for whatever it’s worth, I do like this as an opt-in attribute and think it’s worthwhile. I also don’t really enjoy having huge GitHub debates on these things either so I’ll just leave it at that 😅

matthewp:
@justinfagnani 

> 1. Interoperable incremental hydration across web components. Componentents should not automatically hydrate upon being defined, but wait for a signal from their parent or other coordinator.

Given this component:

```js
customElements.define('hello-world', class HTMLElement {
  constructor() {
    super();
    this.attachShadow({ mode: 'open' });
  }
  connectedCallback() {
    this.shadowRoot.innerHTML = `Hello world`;
  }
});
```

This component does not require any additional configuration from the parent. If this component were to adopt the proposal it would result in worse UX.

justinfagnani:
> The most pressing question for all of these community standards is, imo, "would native elements adopt this"?

I don't think this is true. If these were proposals for standards, then they should be proposed against the standards. These instead are proposals for how userland code, including vanilla elements and libraries, can collaborate and interop. Context and Pending State for instance, are not the type of things we'd proposal to WHATWG.

For community protocols we're looking for protocols that multiple vendors/authors could benefit from using in an interoperable way. We need a common place for the specification for these prototols.

If someone else besides Lit wants to use defer-hydration, they would be able to be adhering the to spec here, but no one is _forced_ to. If you want your elements to hydrate immediately and possibly render multiple times as they get new properties set by hosts? You can obviously do just that and ignore this protocol.

justinfagnani:
Defer hydration by default just enforces a top-down ordering of hydration. Aside from better performance, it shouldn't be noticable to users. How would that be worse DX?

matthewp:
> Defer hydration by default just enforces a top-down ordering of hydration. Aside from better performance, it shouldn't be noticable to users. How would that be worse DX?

The parent is not guaranteed to be defined at the same time as the child. The child is waiting for a parent that could very well not be defined until much later into the future.

```js
import './child.js';

button.onclick = () => {
  import('./parent.js');
};
```

In this example `child.js` is waiting for `parent.js` even though it probably doesn't need to. This means that `child.js` is not interactive when it probably should/could be.

matthewp:
Here's an example of two components, one that honors `defer-hydration` and one that does not: https://codepen.io/matthewp/pen/jOpBbLG?editors=1010

Note that as a web component author you can assume that your code will be loaded in any certain why. Web components are designed to be isolated units. The consumer decides when and how they are loaded. You decide when your element renders/hydrates. You must make an decision based on your knowledge of your components requirements, which defers from component to component.

justinfagnani:
That example is demonstrating that a component can ignore defer-hydration (which is the default, of course) and hydrate as soon as it's defined.

matthewp:
@justinfagnani the draft of this proposal does not state that components should ignore `defer-hydration` by default. It states the opposite:

> 1. Interoperable incremental hydration across web components. Componentents should not automatically hydrate upon being defined, but wait for a signal from their parent or other coordinator.

If you'd like to amend the proposal to state that components should ignore it by default and only honor it in a certain set of circumstances I'd be happy to work with you on what those circumstances should be.

justinfagnani:
I'm confused by the request. Custom elements already ignore unknown attribute by default.

matthewp:
What unknown attribute? This proposal makes `defer-hydration` be known. It suggests that component authors honor it and it's proposed meaning.

justinfagnani:
It's unknown to elements that don't implement this protocol. The whole protocol is opt-in and if it doesn't suit your needs there's no reason to implement it.

matthewp:
All protocols are opt-in. The purpose of this repository is to discuss and debate the merits of the ideas, not to rubber-stamp every one that is brought up.

This proposal explicitly states:

> 1. Interoperable incremental hydration across web components. Componentents should not automatically hydrate upon being defined, but wait for a signal from their parent or other coordinator.

I think this is patently false, for most components. Most components *should not* wait. Most components should be designed without assumptions of how they will be loaded or used.

This decoupled nature is what makes web components unique and special. This proposal makes web components more like traditional, highly coupled, framework components. That's fine if you want to build that way, but it should not be a recommendation for the community.

justinfagnani:
This proposal is not passing judgement on all components. It's not saying all components should implement it. I'll change the wording somehow to state that some components should wait for their parents (yet it's absolutely true that if components get data from their parents in hydration, they should wait for their components to hydrate first). But really, if you don't need this proposal you simply don't have to use it.

matthewp:
"You don't have to use it" isn't a good enough answer. These protocols are seen as recommendations by the community at large. The use-cases need to be specific. Eleventy/webc has already implemented this in a blanket way, which will lead to poor outcomes like the one I provided in the Codepen example.

If components get data from their parent in hydration it should match the data that has already been serialized via HTML attributes and children. Although I'm not sure why the data is passing data down to children that it has already set via attributes? That seems like a bug in the parent component.

There might be some cases where data cannot or should not be serialized in this way. I'm happy to discuss those specific cases and what the solution should be.

Then if we find some cases where a `defer-hydration` is the right answer then I'd be happy if this proposal spelled out the case explicitly. But we must first discuss the use-cases. You can see @kevinpschaaf and I discussing some above and the only one we have found so far are app components that are tightly coupled.

kevinpschaaf:
> There might be some cases where data cannot or should not be serialized in this way. I'm happy to discuss those specific cases and what the solution should be.

To be very clear, the original motivation for this proposal is addressing this exact case, where the initial state cannot be serialized for whatever reason. If the state can be serialized with no compromises, there is no need for this protocol. So _YES_, we're only discussing those specific cases and what the solution should be. Sorry if that was not clear, and we can make the protocol proposal more clear about when it's needed.

The nice thing about this proposal is that it is always safe for a SSR implementation to add `defer-hydration` and for the client-side hydration for that implementation to remove `defer-hydration`, even if the component being rendered didn't implement `defer-hydration`. In that case, the component doesn't observe the attribute, and doesn't care. On the server side, while it's always safe to add the attribute, the only time an SSR implementation would _need_ to add `defer-hydration` is if it is setting non-serializable data (e.g. via a JS property) -- the corollary being that if a component has an API that receives non-serializable data, it wouldn't be able to be efficiently hydrated without implementing `defer-hydration`.

It seems that your concern is that if your component doesn't _need_ `defer-hydration` (because it doesn't have a property API for accepting non-serializable data), it might still be considered a "bad actor" in the community if it doesn't implement the protocol. But as the above paragraph describes, it would only be a "bad actor" if it actually needed `defer-hydration` to efficiently hydrate because it accepts non-serializable data but didn't implement the protocol. And on the rendering side, it would be fine for an SSR implementation to not add `defer-hydration` if it was always able to serialize data to the components it's rendering. OTOH, if the SSR implementation passed JS values down the tree that are used for rendering but are not serialized (e.g. the `<data-tree>` example posited above), that's the point where it's being a bad actor by not using `defer-hydration`, because those components won't upgrade with the server-rendered data.

In this way, the protocol seems to be to be fully opt-in _when you need it_ (both on the server and component sides), and if you don't need it, you're no worse for not implementing it. Where is the stress?

justinfagnani:
> These protocols are seen as recommendations by the community at large.

This is a repository of interoperable protocols, it is very much *not* a repository of recommendations. We are not saying that components _should_ implement these protocols.

People may (and probably do!) disagree with the concept of Context. They don't have to use the context protocol. But if you do want an interoperable way of doing contexts, we have a specification here you can implement to let you work with other components.

Likewise, people who have a need for this kind of deferring can implement this protocol to be interoperable with other components that do.

matthewp:
@kevinpschaaf I'm looking at this from the perspective of a component author. I believe you and Justin are looking at it from the perspective of framework authors. I think this is the source of our disagreement.

> It seems that your concern is that if your component doesn't need defer-hydration (because it doesn't have a property API for accepting non-serializable data), it might still be considered a "bad actor" in the community if it doesn't implement the protocol. 

No, I don't care about that at all. My concern is that components will implement the protocol even if they do not need it, which will lead to bad UX like the Codepen example I showed above.

In fact, the Lit implementation does so, it implements it in a blanket way even if the component doesn't have non-serializable props: https://github.com/lit/lit/blob/761375ac9ef28dd0ba8a1f9363aaf5f0df725205/packages/lit-element/src/experimental-hydrate-support.ts#L69-L71

That means that if the component is run in an environment without Lit's orchestrator it will result in the exact same bug as in the Codepen.

----

And just for the record I don't think I agree that any component actually needs this. So far there has been resistance to getting into specific use-cases so we haven't had a chance to discuss them in depth. I'm happy to get into the specific use-cases if you'd like. 

A component should already have all of the information to know if it has received non-serializable props or not and do the appropriate thing. For ex:

```js
if(this.someProp === undefined) {
  // do nothing
  this.attachEventHandlers();
} else {
  this.render();
  this.attachEventHandlers();
}
```

If there's something I'm missing here I'm happy to dig into it.

matthewp:
@justinfagnani 

> Likewise, people who have a need for this kind of deferring can implement this protocol to be interoperable with other components that do.

Unless the text is more specific on when it is needed people are going to implement it in a blanket way. We now have 2 frameworks, Eleventy and Lit, which implement this proposal even on components that do not need it. So I don't think your prediction is actually true.

We need to get into specific use-cases and if we find them, update the text so that people know *when* they should do this. Otherwise, like Lit and Eleventy, they will think that they should do it always.

justinfagnani:
> No, I don't care about that at all. My concern is that components will implement the protocol even if they do not need it, which will lead to bad UX like the Codepen example I showed above.

I don't quite get this concern. Why would someone implement this protocol if they don't need it? You wouldn't implement context if you don't need context.

And what's the bad behavior in the codepen? You told a component that implements the protocol to wait with `defer-hydration`. That's fine. If you don't want the component to wait, don't add the attribute. It's opt-in from the parent _and_ child sides - even for components that implement the protocol.

The fact that Lit implements this protocol is because the problem of upgrade ordering in SSR conflicting with CSR is pretty pervasive and this is a comprehensive fix - that is essentially unobservable to users. But again, even of the components that implement the protocol the behavior is opt-in. If `defer-hydartion` isn't present, the component hydrates immediately.

> That means that if the component is run in an environment without Lit's orchestrator it will result in the exact same bug as in the Codepen.

No. Lit doesn't have an orchestrator. If the component runs in an environment that didn't set the `defer-hydration` attribute, then the component will hydrate immediately, as is the default.

matthewp:
@justinfagnani The Codepen is showing how loading order can cause a component to render late. In this case `child-two` can render immediately, but doesn't because it implemented this protocol.

Here's another scenario that maybe makes it easier to understand. Let's say you build a Lit app and you have your `hydrate()` code in `main.js`. You have a Counter component in `counter.js`.

Your HTML looks like this:

```html
<script type="module" src="./counter.js"></script>
<script type="module" src="./main.js"></script>

<main>
  <my-app>
    <template shadowroot="open">
      <my-counter defer-hydration></my-counter>
    </template>
  </my-app>
<main>
```

Now let's say that `counter.js` loads in 100ms. `main.js` takes 5 seconds to load because of network congestion.

That means that `<my-counter>` remains non-interactive for almost 5 seconds even though there's no reason for it to be.

Critically, this proposal creates a single point of failure where the top-level "orchestrator" (in the case of lit the thing calling hydrate()) can delay the interactivity of the entire tree if it fails (or is slow) for any reason.

justinfagnani:
Those imports are not realistic, IMO: If `<my-app>` has `<my-counter>` in its shadow DOM, then `my-app.js` should be importing `my-counter.js`.

Even so, deferring `<my-counter>` is _absolutely_ the right default behavior here. `<my-app>` could install an event listener during hydration. If `<my-counter>` hydrates first, it could fire an event that `<my-app>` misses. This is part of the reason why enforcing top-down hydration is so critical. An interactive component like a button very well could hydrate early, but then be seemingly functional but broken as parents aren't ready to have the button be clicked.

If `<my-app>` and `<my-counter>` have special knowledge of each other, such that `<my-app>` knows that it doesn't depend on hydration to correctly use `<my-counter>` (like event listeners), and that `<my-counter>` doesn't depend on hydration of `<my-app>` (like it setting properties on children), _then_ you could safely omit the `defer-hydration` attribute.

This would make `<my-counter>` what we're calling on our team a "hydration root", which can safely ignore `defer-hydration` (or have the parent not set the attribute). There's a specific contract around hydration roots because they can't do anything other components need to observe without waking up those components (loading _and_ hydrating) and possibly buffering events, and they need to serialize all data (ideally without too much duplication, which requires data keys and stores), and those things require protocols too.

> Critically, this proposal creates a single point of failure where the top-level "orchestrator" (in the case of lit the thing calling hydrate()) can delay the interactivity of the entire tree if it fails (or is slow) for any reason.

Again, this is basically the point. We can't know what modules and what hydration steps are critical to the working of the subtree without more information. By default, if the ancestors of the tree are slow to load, we can't just let the children go first without risking lots of bugs. We do assume that the CSR ordering works, so by default we enforce the same hydration order for SSR as rendering order for CSR.

kevinpschaaf:
> In fact, the Lit implementation does so, it implements it in a blanket way even if the component doesn't have non-serializable props

We may very well adjust the heuristic for when `defer-hydration` is used, and/or give users control over when it's used (we already have a flag to control whether top-level CE's are deferred, since we can't know whether they may need to wait for JS data to be assigned before hydrating, or whether they might be fully serialized or fetch data themselves). The current Lit implementation is in "labs" precisely because we know we still have work to tease out the correct patterns here, so I wouldn't over-index too much on that.

That said, in day-to-day life, when building applications we generally recommend setting properties over attributes for passing dynamic data to children (where attributes are more appropriate for static, serializable configuration), since practically speaking, repeated re-serialization of dynamic data is costly and generally of low benefit when you can just pass JS values by reference. Because letting user code that does this run unchanged during SSR has a lot of value in reducing overall complexity (vs. somehow knowing when it's safe to take a user value that was passed via JS in the template and decide to serialize it to an attribute), we've found in practice that this protocol is very easy to implement and lets the user code doing JS-value propagation run unchanged during SSR and efficiently hydrate on the client. Because a tree of components that are intended to hydrate together will generally be bundled together, the upgrade+defer-hydration removal cascade will happen at microtask timing, never resulting in the timing issues you're highlighting.

> I'm looking at this from the perspective of a component author. I believe you and Justin are looking at it from the perspective of framework authors. I think this is the source of our disagreement.

Possibly so. I think the reality is that SSR of a complex tree of components (or even CSR for that matter) can't ever be implemented naively with the expectation of good UX. The example above feels contrived because loading element definitions granularly as separate modules -- while something you are free to do -- is never going to result in optimal DX (for reasons that include network uncertainty, FOUC, ordering non-determinism, etc.) -- and clearly gets even more gnarly when hydration is involved, so using this as a "but what if a user does something naive" point doesn't feel persuasive... Making a robust SSR implementation with good DX and good performance+UX is _exceptionally_ hard and will require following patterns for serialization, loading, and hydration that are known to result in good outcomes. And as much as I'd love it to be possible for a component author to slap together web components from various sources without any cooperation required -- we're only going to reach that good DX/UX bar by having authors and libraries coordinate via protocols and patterns known to scale, and yes at a certain limit that may require help from WC libraries/frameworks.

Our goal here is that we're not defining these patterns and protocols in way that is siloed to Lit -- that would defeat the whole point of Web Components -- but rather in a way that other libraries and even individual component authors can implement them.

kevinpschaaf:
> A component should already have all of the information to know if it has received non-serializable props or not and do the appropriate thing. For ex:
> ```js
> if(this.someProp === undefined) {
>   // do nothing
>   this.attachEventHandlers();
> } else {
>   this.render();
>   this.attachEventHandlers();
> }
> ```

This isn't the case for a component like this:

```js
class MyElement extends HTMLElement {
  someProp = new DefaultThingy('some default value');
  ...
}
```

Because properties can have defaults and be overridden optionally at the call site, it can't generally be known by the component itself when it has received hydration-critical configuration from its host. Consider these cases and how an element would know at upgrade time whether it is fully configured or not:
```html
  <my-element></my-element>
  <my-element .someProp=${new SpecializedThingy()}></my-element>
  <my-element .someProp=${new DefaultThingy('configured differently')}></my-element>
```

Now expand that from one property to many per element; even if you wanted to come up with some regime where you held off hydration based on some diff of properties against known default values (requiring deep diffing, as the 3rd case highlights), that's a whole lot of bookkeeping an element needs to do. Further, ES6 classes don't even give you access to class field default values since constructor initializers run after the `super()` call, making it even harder to build that sort of "is this the default value or not" feature into a WC base class, so you're left to every component doing lot of manual bookkeeping to implement something like you envisioned, or defining class field defaults in a non-idiomatic way.

matthewp:
@justinfagnani 

> Those imports are not realistic, IMO: If <my-app> has <my-counter> in its shadow DOM, then my-app.js should be importing my-counter.js.

This bug is likely to occur any time there are multiple top-level components on the page (which, this is web components, should be all of the time).

If you have components A and B that both use C. If A loads faster than B, then C should hydrate in both roots once A loads. So you've created a scenario where the slowest script holds up your entire application.

I would expect that this bug can and will occur in any app using an islands pattern, for example.

> Even so, deferring <my-counter> is absolutely the right default behavior here. <my-app> could install an event listener during hydration. If <my-counter> hydrates first, it could fire an event that <my-app> misses.

This really feels like you are using a sledgehammer when the right tool is a scalpel.

Missing events is not a bug. Events are not meant to be something that you *have to catch*. Every built-in element that contains events can/will start firing them before your script runs. If an element requires that you hear its events, that's a poorly designed element.

This problem is the same as the "how do you know if an img loaded" problem. The solution is to provide APIs at the element level. With the `` tag you don't "defer hydration" on it so that you can listen to `load`. You check `img.naturalHeight > 0`.

> Again, this is basically the point. We can't know what modules and what hydration steps are critical to the working of the subtree without more information. By default, if the ancestors of the tree are slow to load, we can't just let the children go first without risking lots of bugs. 

I know that when you say "we" you mean your framework, Lit. The application developer *does know* what elements are critical and *they can* handle those elements specially themselves.

I understand that you want to solve this problem at the framework level. I think preventing bugs in poorly designed elements at the sacrifice of gradual / async upgrade -- one of the core benefits of using web components -- is the wrong tradeoff to make. I think you could come up with more targeted APIs for these exceptional cases.

But it's your framework and if you want to prioritize the edge cases you can do so, but we don't need and shouldn't have a community protocol for that.

justinfagnani:
@matthewp 

> This bug is likely to occur any time there are multiple top-level components on the page (which, this is web components, should be all of the time).

The example you showed has the child element in the shadow root:
```html
<main>
  <my-app>
    <template shadowroot="open">
      <my-counter defer-hydration></my-counter>
    </template>
  </my-app>
<main>
```

This structure is more-or-less impossible to get with client-side rendering because the shadow root is created by `<my-app>` itself. There's no way for `<my-app>` to not have at least a chance to be ready for events from,  or provide properties to, `<my-counter>`.

> > Even so, deferring is absolutely the right default behavior here. could install an event listener during hydration. If hydrates first, it could fire an event that misses.
>
> This really feels like you are using a sledgehammer when the right tool is a scalpel.

I will assert that for shadow children, the sledgehammer _is_ the appropriate solution. We are trying to prevent an initialization order that's impossible with client-side-rendering and therefore try to eliminate an entire class of likely hard-to-debug bugs. The cases where a shadow child can safely hydrate before a host are so special that I absolutely think it's reasonable to have developers specifically indicate that it's ok. Like @kevinpschaaf said, we don't quite have the easy-to-use APIs in Lit yet to ignore `defer-hydration`, we will, likely with more structured feedback about when it might be safe to do so.

For _top level_ elements with light-DOM children, I agree that there are a lot more cases where it's safe for children to hydrate first. And in general, you can't assume a particular upgrade order with client-side-rendering in these cases, so elements already have to be much more robust to various upgrade and event-firing orders. It's also extremely uncommon for a ancestor to set properties on a light-DOM child.

For this and other reasons we do not add `defer-hydration` to top-level elements in Lit SSR.

If your example is rewritten to make `<my-counter>` a light DOM child, the output of Lit SSR would be:

```html
<main>
  <my-app>
    <my-counter></my-counter>
  </my-app>
<main>
```

Note, there's no `defer-hydration` attribute here (there may be _inside_ the shadow roots, again for correct initialization order) and elements will upgrade in whatever order their definitions are registered - just like for client-side rendering of this same structure.

> But it's your framework and if you want to prioritize the edge cases you can do so, but we don't need and shouldn't have a community protocol for that.

This is the wrong way to think about a community protocol. Community protocols are for interop among those who are _choosing_ to interoperate. If we didn't host this here, we'd host it somewhere else. This repo is a collection of possible protocols that elements and libraries can use and we already have interoperable implementations of context and defer-hydration between <is-land> and Lit. That's great and just because you don't want it or disagree with our approach doesn't mean that we should haven't a place for the rest of the community to collaborate. These are not recommendations.

And to address your concern and `<is-land>`, I think that is misplaced as well. The whole point of `<is-land>` is to defer the loading and hydration of a subtree. It's previous behavior was to rename every element to a random name to prevent them from upgrading. _That_ was a sledgehammer and prevented elements from having any option to do work "early", forced a lot of DOM mutations, changed element identity, and more.

`<is-land>`'s previous output for your example would have been something like:

```html
<main>
  <is-land-x-my-app>
    <s-land-x-my-counter></s-land-x-my-counter>
  </s-land-x-my-app>
<main>
```

And when `<is-land>` decided to hydrate that subtree, it would rewrite the DOM without prefixed.

`<is-land>`'s output is now more like:

```html
<main>
  <my-app defer-hydration>
    <my-counter defer-hydration></my-counter>
  </my-app>
<main>
```

Which gives the elements a chance to defer their own hydration _if_ they choose to, and with their _own_ strategies for what work to do before and after hydration, or _ignore_ the attribute altogether, which is the default behavior. `defer-hydration` is just a signal, and components do with it what they want.

This should be a lot better for your concerns. You could be an element vendor who always makes early-hydration-safe components and can just ignore the existence of `defer-hydration`. That's much more control than with an orchestrator disabling your component definition completely.

You can also now disagree with `<is-land>`'s choice to signal to a whole subtree to not hydrate, separately from other uses of `defer-hydration` like delaying shadow children hydration, and simply not use `<is-land>`. Use and mechanism are separated concerns now, linked via a common signal.

You seem to be arguing that we shouldn't define a common signal for this at all, which is your prerogative, I just have to disagree strongly.

matthewp:
@justinfagnani My example wasn't using light DOM. Here's the point I was making, more clearly:

```html
<script type="module" src="one.js"></script>
<script type="module" src="two.js"></script>

<x-one>
  <template shadowroot="open">
    <x-three defer-hydration><x-three>
  </template>
<x-one>
<x-two>
  <template shadowroot="open">
    <x-three defer-hydration><x-three>
  </template>
<x-two>
```

This is server rendered HTML.

Since one and two can/will load at different times, `<x-three>` will be available to upgrade whenever the first one loads. It doesn't have to wait for its ancestor. 

Here's a Codepen that simulates this bug: https://codepen.io/matthewp/pen/oNMwdJj?editors=1010

----

> I will assert that for shadow children, the sledgehammer is the appropriate solution. We are trying to prevent an initialization order that's impossible with client-side-rendering and therefore try to eliminate an entire class of likely hard-to-debug bugs.

This is not impossible client-side. You don't have to import your dependencies in your web component code. It's unconventional not to, but it's not impossible. I know you know this I just felt I need to clarify here in case anyone else reading it doesn't know. Web component definition (ie using `customElement.define`) and usage (ie in your HTML) are decoupled, which is one of the things that makes web components such a unique and powerful a tool.

justinfagnani:
@matthewp 

> This is not impossible client-side.

Can you show me how you'd make `<x-three>` initialize before its shadow host in anything close to a typical scenario? Even if you don't load a component definition, it'll at least upgrade after it's created its shadow host, which had to be upgrade to create its own shadow root. In CSR, without shenanigans like moving nodes between shadow roots, shadow trees are initialized in top-down ordering, regardless of element definition import order.

This case seems perfectly fine with how Lit uses `defer-hydration`. We know that with CSR `<x-three>` would not have been created before its shadow host, and in general could depend on properties from the shadow host or the host could depend on events from it. Deferring it's hydration until the host hydrates is appropriate.

I don't see any bug.

I _do_ see a lots of potential bugs and performance problems if by SSR'ing the structure above, the constructor/connectedCallback order is completely inverted from CSR order.

matthewp:
@justinfagnani This is why we need to talk about concrete use-cases. `<x-three>` *does not* depend on properties from its ancestor. It *does not* emit events that its ancestors must see. This component does not do those things, so this is a bug and unwanted behavior.

This is a fake component so let's talk about some real ones. What are these elements who do have these tight couplings and what are the possible solutions? `defer-hydration` is one possible solution but not the only one.

justinfagnani:
> this is a bug and unwanted behavior.

What is the bugs exactly? That we reproduced the CSR order? What's the observable difference in behavior besides preventing a class of bugs?

We write and help others write tons of components that pass properties to and receive events from shadow children. It's exceedingly common. I don't have analysis stats handy, but I'd say it's far more common than not.

If we want to talk use cases, I'd ask what real-world construction results in the problem you're claiming. Because it seems to require:
1. A host component that renders, but does not import, a child component
2. That then has some other actor in the page import that child component definition
3. Doesn't have any sort of dependency between the child component and the host
4. The child can do something useful without its shadow host being hydrated first

Can you point to a setup like that? And if you can, why would it be using `defer-hydration`?

I guess you could argue that Lit should assume that parent-child shadow components have no dependencies between each other, and so not put `defer-hydration` on shadow children during SSR, but that's not what we see, that would be an argument against Lit's _application_ of `defer-hydration` not the existence of a common signal, and I'd respectfully disagree.

For an example of a hierarchical component see Shoelace's [color picker](https://github.com/shoelace-style/shoelace/blob/fec7ef17aa09ae20dd769165e2b31ec8106f3de8/src/components/color-picker/color-picker.ts#L958) which imports and uses at least 6 component children with various attribute, property, and event bindings between them. This is a design system component. Application components tend to have even more complex bindings. What bug is present from hydrating the parents first?

matthewp:
> What is the bugs exactly? That we reproduced the CSR order? What's the observable difference in behavior besides preventing a class of bugs?

The observation difference is that the `<x-three>` is not visible until 5 seconds after it is defined, even though it has no reason not to be. This means that the user isn't able to interact with it. It could be something like an Add to Cart button and the 5 seconds is the difference between a sale and not. 

> If we want to talk use cases, I'd ask what real-world construction results in the problem you're claiming. Because it seems to require:

It doesn't require any of those things. The codepen perfectly shows it. Both `one.js` and `two.js` depend on `three.js`. It just happens that `two.js` loads more slowly, so its `<x-three>` has to wait.

This doesn't rely on any unconventional programming style.

> We write and help others write tons of components that pass properties to and receive events from shadow children. It's exceedingly common. I don't have analysis stats handy, but I'd say it's far more common than not.

Components that pass properties and receive events don't need this sort of solution. It's only "needed" if:

- In the case of events, the parent *must* receive all events. We need to talk about why they can't miss events. What do you do in the case of `` onload? You're going to miss that one if you render it to HTML.
- In the case of properties, the component does not serialize them as attributes. 

It seems like an implicit goal of this is to somehow abstract away the difference between components written purely for client-side usage and components that might be upgraded from DSD, and that's where this idea is coming from.

But I don't think that's a valid goal; we should only promote patterns that lead to good components. It's a framework job's to make it easier to build good components. Not to cover for you so that you don't have to.

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+
+  attributeChangedCallback(name, oldValue, newValue) {
+    if (name === 'defer-hydration' && newValue === null) {
+      this._hydrate();
+    }
+  }
+
+  _hydrate() {
+    // do template hydrate work
+    // ...
```
nolanlawson, line 80: One thing that's not clear to me from this proposal: what does "hydration" mean? Just `connectedCallback`? Or `constructor` plus `connectedCallback`?

I'm assuming that it only means `connectedCallback`, but you could imagine that a web component could have some logic in its `constructor` that assumes client-side rendering, which assumes top-down rendering, and thus breaks when `customElements.define` is called out-of-order.

Is this situation considered out-of-scope for this proposal?

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+  </head>
+  <body>
+    <x-nav>
+      <template shadowroot="open">
+        <x-header defer-hydration>
+          <template shadowroot="open">
+            <h1>Example</h1>
+          </template>
+        </header>
+      </template>
```
renoirb, line 48: I'm a bit confused about why `defer-hydration` be inside a `template`. 

Aren't template "*inaccessible*" (there's a term, can't find it in memory right now. "*folded*"?). Can this also work on any `HTMLElement`?

(Back about template) Maybe it's part of [*declarative Shadow DOM*](https://developer.chrome.com/articles/declarative-shadow-dom/) (that was unknown to me until writing this); is this limited only to `template[shadowroot=open]` (and not on any other HTMLElement)?  I'm assuming it's meant for components that are returned from a server (including a SSR like Nuxt's) and (for a reason or another) allow adding later on client-side the elements' data (in one way or another), then removing the closest to root ` [defer-hydration]` in `template [defer-hydration]` (or *must it be* only when `template[shadowroot=open]  [defer-hydration]`).

Is that the intended scenario?  I'm not sure I could see that from what I'm reading here.

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
...
+    <script type="module" src="./app.js"></script>
+  </head>
+  <body>
+    <x-nav>
+      <template shadowroot="open">
+        <x-header defer-hydration>
+          <template shadowroot="open">
+            <h1>Example</h1>
+          </template>
+        </header>
```
renoirb, line 47:
```suggestion
        </x-header>
```

proposals/defer-hydration.md
nolanlawson: Thanks for the clarification. It might be good to nail down some SHOULDs in this case (not necessarily MUSTs). E.g., thoughts that pop into my head are:

- Should a component-author-author'ed (as opposed to framework-author-authored) `connectedCallback` fire in a deferred way?
- What about framework-specific callbacks related to reactivity/rendering? E.g. in LWC we have `renderedCallback` for when the DOM is rendered, Lit has various callbacks like `updated`/`firstUpdated`/etc.
- What if somehow a component is disconnected before the deferred `connectedCallback` code fires? Should `disconnectedCallback` fire before `connectedCallback`, or not at all?

proposals/defer-hydration.md
```diff
@@ -0,0 +1,90 @@
+# Defer Hydration Protocol
+
+An open protocol for controlling hydration on the client.
+
+Author: Justin Fagnani
+
+Status: Draft
```
justinfagnani, line 7:
```suggestion
Status: Proposal
```

Westbrook (approved):
Based on the WCCG meeting on 7 November 2023:

Updating this PR to target "proposal" status in all locations will unlock it for merging.

---

We really appreciate the input put into this PR to date and want to remind participants that "community protocols" are not a requirement (even when they move to "Accepted" status) but a suggestion for easier interoperability between components and tooling. In many cases, a competing proposal can be the best path to seeing the changes in the community that you might deem more important than currently available protocols.

For any conversation that various commenters do not feel this PR has fully resolved, please open issues or a PR with changes you'd like to see in the protocol for further, focused review.

matthewp:
I think merging this is a huge mistake. I have disagreed with some of the proposals in this repo but this is the only one that I feel is actually harmful. As stated above, this proposals goal is to abstract away the difference between client-side rendering and hydrated rendering. It does so by making hydrated rendering slower.

It's perfectly fine for a framework to chose this trade-off, but making it a recommendation for the community to also do so, is harmful.

Westbrook:
As a reminder the "Proposal" status level that is suggested as the next step for this conversation is outlined as follows:

> _Proposal_ "Proposal" status applies to just about anything that community members are interested in putting some thought into. While an issue submitted to this repo can help generate initial ideas on the protocol or space of interest and clarify the information needed to kick off a more fruitful conversation, a PR will serve to make known that you are interested in support to drive the protocol in question forward. Having an initial explainer included in this PR, while adding the protocol to the "Proposals" table shown above, will prepare the community to both communicate about and contribute to the development of the protocol asynchronously.

For anyone with continuing concerns about the current shape of the proposal, leveraging issues or PRs for modifications is a great way to follow up on those, or if your concern is even deeper, then you might find submitting an alternate proposal that you felt addressed the concepts herein in a more palatable manner productive to your cause.

justinfagnani:
@matthewp 

> It's perfectly fine for a framework to chose this trade-off, but making it a recommendation for the community to also do so, is harmful.

For frameworks to choose this trade-off they need a specification to interop over. This is that specification. It's entirely opt-in, and once again the `defer-hydration` attribute is only a signal. Elements can do with it what they wish.

We've determined on the Lit project that this feature is required for proper hydration, and will use this feature. By publishing a protocol proposal here we are making it possible for others to interop with us in a formalized way. That is the point of community protocols.

I **strongly** disagree with your assessment of this protocol, but regardless of that there is room in community protocols for protocols that not everyone agrees with. There's even room for competing protocols. As long as they can be named, referenced, and are specified enough for multiple independent and compatible implementations this repository is serving a good purpose.

proposals/defer-hydration.md
justinfagnani: This is a declarative shadow root.

matthewp:
> By publishing a protocol proposal here we are making it possible for others to interop with us in a formalized way. That is the point of community protocols.

That's not the point, if that was the case then everything would get merged. If this is truly the case then I have at least 1 proposal of my own that I'll be getting merged soon.

But I don't think that you think this is the point of the repository, as there are no non-Lit involved proposals that have ever been merged.

justinfagnani:
@matthewp 

> That's not the point, if that was the case then everything would get merged. 

I think there _should_ be a high bar for just blocking a proposal, especially because say that there needs to be two independent implementations to move to "accepted".

You seem to want a veto here, but we haven't established anything like that in the process. Must we satisfy your objections to even make a _Proposal_ even though we've made it very clear our rationale and the problems we _already_ solve with this protocol?

> If this is truly the case then I have at least 1 proposal of my own that I'll be getting merged soon.

Given your tone in this conversation, would you actually be doing this in good faith? Do you use the protocol today? Are you truly looking to interop with other components and libraries? Would you show up to WCCG meetings to discuss it? Because if so, then make the proposal! If you're just trying to gum up the works I think your energy would be better spent elsewhere.

nolanlawson:
I haven't fully grokked the details of this proposal, but we (LWC) will probably be evaluating whether to implement it in the next couple months or so.

I'd be really interested to understand better the concerns around this proposal, and I'm also still a little unclear about what would "count" as a framework-interoperable implementation (e.g. see [my question on nitty-gritty details](https://github.com/webcomponents-cg/community-protocols/pull/15#discussion_r1341850345)).

Maybe this would be worth a breakout session for the WCCG?

mattlucock:
I overlooked this in my previous comment, but components have to opt-in to hydration in the first place, do they not? This might be a silly question, but how do we actually know whether we can safely SSR any given component? Is SSR interoperable? (And if not, how can `defer-hydration` be?)

mattlucock:
I think I was a bit confused when reading the proposal a few weeks ago—although I think that's because the proposal is confusing. Reading it again, it's not apparent that a parent 'asks' its children to defer hydration like I thought; it more seems like a component would determine for itself that it should defer hydration when being SSR-d. But this actually just creates a different variation of the same problem: how does a child know whether its parent participates in the protocol and will remove the `defer-hydration` attribute from it? If the parent never removes the attribute, the child never hydrates.

@matthewp It makes complete sense to me that a framework for SSR-ing web components could opt all components in to something like this by default; this thing would basically just be an implementation detail of components built using the framework. But as I was getting at, it feels like this only makes sense in the context of a particular framework, and it feels like this fundamentally cannot be made interoperable in any way that makes sense.

I am quite concerned by the approach that seems to have been taken thus far with these proposals, which appears to be "we've implemented this in Lit, and now we're making this a community protocol so anyone can choose to interoperate with us if they want to". Not only does it presuppose that Lit's design decisions are the right ones and should determine the design of components broadly, but it also presupposes that these things can and should be interoperated with.

nolanlawson:
I'm actually glad that @justinfagnani put the effort into documenting how Lit does things, so that other frameworks might be able to interoperate with Lit. This is kind of the whole goal of web components.

Thinking more about this proposal, though, I feel like it may be putting the cart before the horse – before figuring out how 2 WC frameworks should hydrate together, we need to figure out how they'll _SSR_ together. E.g. if LWC wanted to implement support for SSR'ing Lit components and hydrating with them, it's not enough to use `defer-hydration` correctly – we'd also have to figure out some generic way to SSR a WC (and ideally use some kind of standardized abstraction layer and not just directly import `@lit-labs/ssr`).

Again, I'm not against this proposal, but I'd like to see some brave WC framework author figure out how to SSR/hydrate Lit components, and then it'd be clearer to me how this is supposed to work in practice.

Alternatively: if this proposal could be distilled down into something non-hydration related (e.g. `defer-work` or `defer-rendering`, etc.), then maybe it could serve as the low-level primitive that the more complex system could leverage.

mattlucock:
@nolanlawson The point I was trying to make is that "the whole goal of web components" is not, and cannot be, to "interoperate with Lit". I think I may be over-emphasizing this concern, but this framing and thinking is really unproductive for what is supposed to be a community initiative, and I also think it's backwards: it's not that web components should interoperate with Lit, but that Lit should interoperate with web components.

I think I generally agree with the rest of your sentiment, but I am a bit pessimistic about there being a distilled form of this that can be truly interoperable. There is nuance to this, which I mentioned in my previous comment, in that this seeks to apply behavior to a component from the outside, only some components will support it, and if we try and apply it to a component that doesn't support it we will draw incorrect conclusions about that component's behavior.

nolanlawson:
@mattlucock These are valid concerns. The goal here is definitely interop among multiple WC frameworks – not just Lit. However, Lit might be at the vanguard of figuring out this whole SSR/hydration thing, so I take their proposal seriously when they say that something like `defer-hydration` is a useful primitive.

If it's not, then by all means, let's build something better! But I respect Lit for coming to this forum to at least reach out to other WC frameworks and try to figure out how to interoperate.

nolanlawson:
@dgp1130 I think what you're running into is called a "node reaction callback" error. It should be catchable with an [error listener on the `window`](https://nolanlawson.com/2023/08/25/catching-errors-thrown-from-connectedcallback/).

I've found this to be a generic problem any time you're trying to catch an error in `connectedCallback`, form association callbacks, etc. The best solution I've found is the global `window` listener.

dgp1130:
@nolanlawson Thanks for that context and excellent blog post. I can definitely understand why `attributeChangedCallback` works that way. It does make some amount of sense that `el.setAttribute('attr', 'value')` shouldn't ever fail. But it is a challenge for this proposal.

Would it be feasible to use the `window` `error` event to catch an uncaught `attributeChangedCallback` error and be confident that was the source of the error? I'm not convinced. [MDN states:](https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event)

> The `error` event is fired on a [`Window`](https://developer.mozilla.org/en-US/docs/Web/API/Window) object when a resource failed to load or couldn't be used — for example if a script has an execution error.

So more than just uncaught callback reaction errors will trigger it. Is there a risk of potentially catching errors which did not originate from an `attributeChangedCallback` handling the `defer-hydration` removal?

Fortunately `attributeChangedCallback` is synchronous, so resource load error or script execution error shouldn't be an issue in this context. However, there are three cases I can think of where listening to `window` `error` would catch incorrect errors.

First is that component or its dependencies may call `window.dispatchEvent('error')` at any time. I can't say I've seen this in practice too much but I could see some scripts essentially throwing an async error via this mechanism.

Second is composed hydration. Take the following example of hydration "call stack":

```
<outer-component>
  <inner-component defer-hydration>
    <error-component defer-hydration></error-component>
  </inner-component>
</outer-component>
```

In this case, we're using `defer-hydration` to create top-down hydration timing. `outer-component` hydration triggers `inner-component` hydration which triggers `error-component` hydration.

`error-component` throws an error on hydration in `attributeChangedCallback`. Both `outer-component` and `inner-component` have listeners for `window` `error`, but they cannot distinguish whether the error is coming from a direct descendent. In a "real" call stack, an error from `error-component` could be caught and ignored by `inner-component`, preventing `outer-component` from ever seeing it. But with this event listener, both `inner-component` *and* `outer-component` observe the error. There is no way for `outer-component` to know that the error was caught and handled by `inner-component`.

[Example Stackblitz](https://stackblitz.com/edit/stackblitz-starters-j8fhu4?description=HTML/CSS/JS%20Starter&file=script.js,index.html&terminalHeight=10&title=Static%20Starter)

Third is catching errors from `setAttribute` or `connectedCallback` invocations unrelated to hydration. Take for example:

```typescript
class MyComponent extends HTMLElement {
  static observedAttributes = ['defer-hydration', 'attr'];

  attributeChangedCallback(name, _oldValue, newValue) {
    if (name === 'defer-hydration' && newValue === null) {
      this.onHydration();
    } else {
      throw new Error('Unknown attribute'); // Caught by `window` `error` callback.
    }
  }

  onHydration() {
    this.setAttribute('attr', '');
    console.log('Hydration successful!'); // Logs as expected.
  }
}
```

That `setAttribute` call can produce a `window` `error` event if `attr` happens to be observed by the component and an error is thrown in `attributeChangedCallback`. However `attributeChangedCallback` did not throw in the `defer-hydration` case, which is the error we'd actually want to catch. The unrelated error when modifying `attr` is swallowed and shouldn't fail hydration.

[Stackblitz demo](https://stackblitz.com/edit/stackblitz-starters-dijkv6?description=HTML/CSS/JS%20Starter&file=script.js,index.html&terminalHeight=10&title=Static%20Starter)

One other challenge is that multiple `attributeChangedCallback` or `connectedCallback` errors might result in multiple errors caught by `window` `error`. I suspect it would be slightly more correct to construct an `AggregateError` of each caught error. Alternatively, the *last* error might be the most important because it is likely the one which was actually thrown in `attributeChangedCallback`. Even that's not perfect however, as it ignores the use case of one component catching and ignoring the error from another component.

Does `window` `error` give more information about the events it emits such that we could distinguish some of these cases? My guess is no, but documentation on this seems a little lacking. My intuition is that using `window` `error` for hydration errors in this context would be too inconsistent and ambiguous to be generally useful. Would love to hear others' thoughts.

nolanlawson:
> Would it be feasible to use the `window` `error` event to catch an uncaught `attributeChangedCallback` error and be confident that was the source of the error?

If you set the global `error` listener right before calling `setAttribute`, and if you check `event.error.stack` to see if it contains `attributeChangedCallback`, then you could be reasonably certain. But it'd still be kind of janky.

dgp1130:
Wouldn't that still fail for the third example I gave? Any error in an `attributeChangedCallback` will have that method in the stack trace. We'd need to identify that it was `ErrorComponent.prototype.attributeChangedCallback` specifically. That sounds tricky with minifiers (though potentially possible). But I think it would fail in the case where there are two `ErrorComponent` classes since they would not be distinguishable. Beyond that you could have two instances of an `ErrorComponent`, which one we want to catch depends on which one I was hydrating, which I don't think is knowable by looking at stack traces.

nolanlawson:
@dgp1130 IIUC yeah, if you have an `attributeChangedCallback` that is itself invoking another `setAttribute` whose `attributeChangedCallback` throws an error, then yeah, I don't see how you could reliably distinguish the two errors. It would have to be part of the `defer-hydration` contract to throw a particular type of error or something.

Westbrook (commented):
We agreed as per the March monthly WCCG meeting that merging this PR is appropriate, being it targets the "Proposal" status, the lowest possible level. We'll do our best to capture any conversations into issues for follow-up as the protocol moves up the various status levels (or not), so feel free to get your last comments posted over the next week and a half. On the 25th of March this PR will be merged as a "Proposal" protocol.

README.md
```diff
@@ -13,10 +13,12 @@ Check out the [Issues](https://github.com/webcomponents/community-protocols/issu
 | Proposal       | Author           | Status     |
 |----------------|------------------|------------|
 | [Context]      | Benjamin Delarre | Candidate  |
+| [Defer Hydration] | Justin Fagnani | Proposal |
 | [Pending Task] | Justin Fagnani   | Draft      |
 | [Slottable Request] | Kevin Schaaf   | Proposal  |
 
 [Context]: https://github.com/webcomponents/community-protocols/blob/main/proposals/context.md
+[Context]: https://github.com/webcomponents/community-protocols/blob/main/proposals/defer-hydration.md
```
Westbrook, line 21:
```suggestion
[Defer Hydration]: https://github.com/webcomponents/community-protocols/blob/main/proposals/defer-hydration.md
```

PR #23, opened 2021-09-01
Require implementations to move to "Accepted"

matthewp:
This changes the language to require 2 implementations before moving to "Accepted".

This is intended to prevent accepting a protocol that no one objects to, but only 1 party is interested in implementing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,12 @@ Community Protocols will go through a number of phases during which they can gat
   A protocol generally agreed to be an interesting area of investigation is given "Draft" status. At this point, the conversation will pivot away from proving the need for a protocol and towards proving a specific pattern by which the protocol can be achieved across the web components community. Developmental work in this area can be addressed in PRs or via group meetings of the [w3c's Web Components Community Group](https://github.com/w3c/webcomponents-cg) as needed.
 
 - *Candidate*
-  Once a protocol has received proper incubation and a pattern for applying it has solidified it will be given "Candidate" status. This status outlines that it will soon be accepted. Community members should use this opportunity to outline any final concerns or adjustments they'd like to see in the protocol before adoption. By this point, the "work" should mostly be done and the focus will be polish and presentation.
+  Once a protocol has received proper incubation and a pattern for applying it has solidified it will be given "Candidate" status. This status outlines that it will soon be accepted. Community members should use this opportunity to outline any final concerns or adjustments they'd like to see in the protocol before adoption. By this point, the work should mostly be done and the focus will be polish, presentation, and implementation.
 
 - *Accepted*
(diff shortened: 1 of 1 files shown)


Westbrook (approved):
LGTM!

thescientist13 (approved):

README.md
```diff
@@ -29,11 +29,13 @@ Community Protocols will go through a number of phases during which they can gat
...
+  Once a protocol has received proper incubation and a pattern for applying it has solidified it will be given "Candidate" status. This status outlines that it will soon be accepted. Community members should use this opportunity to outline any final concerns or adjustments they'd like to see in the protocol before adoption. By this point, the work should mostly be done and the focus will be polish, presentation, and implementation.
 
 - *Accepted*
   An "Accepted" protocol is one that is ready to put into action across the community. With clear APIs, types, usage patterns included they are ready to serve the greater purpose of supporting interoperability between web components built from various contexts.
 
+  To move to "Accepted" a proposal needs __2__ implementations.
```
graynorton, line 37: Given the nascent nature of the web components tools & libraries space, the WCCG and the process for common protocols, it feels a bit premature to state this requirement so concretely.

How about generalizing the language for the time being to capture the spirit of the requirement while acknowledging that the criteria and the governing process still need to be fleshed out? Maybe something like this:

> While the formal process for moving proposals through the various stages is not yet specified, it is generally expected that a proposal will not move to "Accepted" until it has at least 2 implementations.

matthewp: The point of the PR is to establish a requirement, so I'm not sure there would be much value to the addition if it read like a suggestion.

I think a PR is a perfect place to establish such process. And even though the requirement is concrete, that doesn't mean the requirement can't be amended or removed in the future via the same process.

justinfagnani: I think this is ok to go in considering that this is a changeable process - we're defining it and we can tweak it as we go and get experience with thing.

One thing that immediately come up when discussing implementations is what counts as an implementation. I could see multiple component libraries all use a common helper library to implement a protocol. That indicates independent multi-party interest, but might not technically be multiple implementations.

I do think there's more than one way to approach the "only one-party's interested" problem:
* "Acceptance" could require positive approval from more than one member (in the W3C sense - ie, an organization).
* "Acceptance" might mean that the CG as a whole - possibly with consensus - wants to see a protocol be implemented and we add an additional phase which indicates there's multiple compatible implementations.

Anyway, I think this fine and we have +1s from a few people, but I do think we'll need to refine this even more.

matthewp: I think multi libraries adopting the feature, even if using a common libraries, counts as multiple implementations. Web browsers have in the past used common libraries between them and this is essentially the same. As long as the protocol as written *can* be implemented independently (it doesn't say "you must use this library", ala what happened with websql).

PR #31, opened 2022-04-21
fix: clarify ordering of stopPropagation / callback

nolanlawson:
As discussed in the community group meeting today, this clarifies the ordering of `event.stopPropagation` and `event.callback` to ensure propagation is stopped even if the callback throws an error.


benjamind (approved):
LGTM!

thescientist13 (approved):

Westbrook:
Thanks @nolanlawson!!

justinfagnani:
btw, another way to express this is that the provider must call stopPropagation() even in the presence of errors in the consumer callback. That doesn't necessarily require a specific order of the calls. A caution provider will probably want to wrap the callback in a try/catch block.

PR #34, opened 2022-05-18
fix typo in context proposal

keithamus:
I believe this is meant to be `loggerDisposer()` - disposing of the old provider, to start using the new one.

### DIFF
--- a/proposals/context.md
+++ b/proposals/context.md
@@ -172,7 +172,7 @@ class SimpleElement extends HTMLElement {
         (value, dispose) => {
           // protect against changing providers
           if (dispose && dispose !== this.loggerDisposer) {
-            this.dispose();
+            this.loggerDisposer();
           }
           this.logger = value;
           this.loggerDisposer = dispose;


proposals/context.md
```diff
@@ -172,7 +172,7 @@ class SimpleElement extends HTMLElement {
         (value, dispose) => {
           // protect against changing providers
           if (dispose && dispose !== this.loggerDisposer) {
-            this.dispose();
+            this.loggerDisposer();
```
thescientist13, line 175: I'm admittedly not too familiar with this proposal, but looking at the previous example in the doc, I _think_ the change would be to just drop `this`?
```js
if (dispose && dispose !== this.loggerDisposer) {
  dispose()
}
```

justinfagnani: `this.loggerDispose()` is what we want here since we're keeping the new value - we want to dispose the _old_ context provider. If we didn't have `this.logger = value` on like 177, then we would want to call `dispose()` (and we could do it earlier like you say).

justinfagnani (approved):

proposals/context.md
thescientist13: Awesome, thanks for clarifying.  Makes sense now.

PR #36, opened 2022-10-28
[context] Replace context object interface description to allow any type as a context

### LINKED ISSUE #19: [context] are context objects needed?
Reading the Context proposal one thing that sticks out to me is that it's a bit unclear what the purpose of the Context object is. It seems like it's used by a provider as a filter for contexts that it can supply.  Assuming that's the case, wouldn't make the type of context being requested part of the event name instead?

ie, instead of `"context-request"` it becomes `"context-request-theme"` and a theme provider listens to that specific event rather than all context events.

justinfagnani:
Fixes #19 and aligns with Lit's implementation.

cc @benjamind @EisenbergEffect

### DIFF
--- a/proposals/context.md
+++ b/proposals/context.md
@@ -32,7 +32,9 @@ The Context API does not intend to cover all cases and forms of Dependency Injec
 
 **Context API is not a state management alternative**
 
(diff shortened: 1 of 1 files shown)


Westbrook (approved):
LGTM.

benjamind:
LGTM!

Issue #21, opened 2021-08-25
[context] do providers retain references to consuming components?


benjamind:
Moving [this conversation](https://github.com/webcomponents/community-protocols/issues/19#issuecomment-905135476) here.

> My question is still related about the Provider. Its lifecycle, how we should work when we're implementing one. I see @benjamind comment made on July 7th makes sense.

> What feels unclear is if we have to have the Provider to hang on to that context-request Event reference, and make it so that it uses it when it detects it's time to disconnect

benjamind:
I think you have the right idea, though I think this is off-topic to the issue here.

It may be helpful to check out the lit-context implementation I've started [here](https://github.com/lit/lit/pull/1955). Especially this [ContextProvider](https://github.com/lit/lit/pull/1955/files#diff-53fa6c8f234974eb9d4a8421925e380d5b2a34741a7e134c9328f722513d8021R38) implementation.

Assuming you wish to create a Context Provider which can deliver the context more than once, i.e. when its context value changes, will have to hold onto the callback function which is provided in the ContextEvent for any context it is interested in satisfying, which _also_ has the `multiple` property in the event set to true (indicating its willing to receive the value more than once). It shouldn't retain the Event object, only the callback from within it. Whenever the context value in the provider changes (not properties on the context value, the value reference itself), the provider should call the callbacks it has collected again.

One thing that I think is wrong in your implementation here is that your provider is driving the `requestUpdate` call on the consuming component. This I think is violating an encapsulation rule. The context provider should know nothing about the rendering lifecycle of a context consuming component, it should merely provide the value in the callback, and leave updating to the consumer.

renoirb:
> **renoirb**: What feels unclear is if we have to have the Provider to hang on to that **context-request** Event reference, and make it so that it uses it when it detects it's time to disconnect.  

(Added after)

Like [a "*Stateful Context Manager*"; knowing which ContextAPI components are currently live, are accepting updates, etc.](https://gist.github.com/renoirb/21e31aab8d4cbcebb24afede7c49e449#file-context-api-ts-L85-L147)

Not necessarily to do updates, like we see below with `requestUpdate` (yes it is violation — That is copy-pasta from experiments from a month ago). But more about an utility to help track/untrack that can be used to help orchestrate consumer/provier operations or lifecycles.

```ts
/**
 * In experiment above, this was called UpdatableHonk — Naming things are hard.
 * Pretty much an event with a target property.
 */
[...]

PR #40, opened 2023-02-02
[context] Rename `multiple` to `subscribe`

### LINKED ISSUE #24: [context] improve naming of dispose and multiple arguments
Currently the proposal outlines that the `context-event` should carry a payload that can include the optional argument `multiple` in order to indicate that the consumer will handle multiple repeated deliveries of the context value. The callback in the payload then receives an optional second argument `dispose` which is a function which can be invoked by a component to indicate to its provider that it no longer wishes to receive updates.

I'd like to propose that we rename these two parameters to make them more related to each other:

```ts
export type ContextCallback<ValueType> = (
  value: ValueType,
  unsubscribe?: () => void
) => void;

export class ContextEvent<T extends ContextKey<unknown>> extends Event {
  public constructor(
    public readonly context: T,
    public readonly callback: ContextCallback<ContextType<T>>,
    public readonly subscribe?: boolean
  ) {
    super('context-request', {bubbles: true, composed: true});
  }
}
```

So an event is emitted with an argument in its payload to `subscribe` to further updates to the context value, and then receives in its callback a second argument to `unsubscribe` from those updates.

This feels to me like it clarifies the intent and behavior here.

justinfagnani:
Fixes #24 and #21

Clarifies the language around when a provider should keep a reference to a request callback. Some minor related example cleanup too.

### DIFF
--- a/proposals/context.md
+++ b/proposals/context.md
@@ -140,7 +140,7 @@ this.dispatchEvent(
 
 If a provider listening for this event can provide the requested context it will invoke the callback passed in the payload of the event. The element can then do whatever it wishes with this value.
 
-It may also be the case that a provider can retain a reference to this callback, and can then invoke the callback multiple times. In this case providers should pass a unsubscribe function as a second argument to the callback to allow consumers to inform the provider that it should no longer update the element, and should dispose of the callback.
+It may also be the case that a provider can retain a reference to this callback, and can then invoke the callback multiple times. In this case providers should pass an unsubscribe function as a second argument to the callback to allow consumers to inform the provider that it should no longer update the element, and should dispose of the callback.
 
 An element may also provide a `subscribe` boolean on the event detail to indicate that it is interested in receiving updates to the value.
 
@@ -300,7 +300,7 @@ export type ContextCallback<ValueType> = (
  * satisfy the request, calling the `callback` with the requested value if so.
  *
  * If the requested context event contains a truthy `subscribe` value, then a provider can call the callback
- * multiple times if the value is changed, if this is the case the provider should pass a `unsubscribe`
+ * multiple times if the value is changed, if this is the case the provider should pass an `unsubscribe`
  * function to the callback which requesters can invoke to indicate they no longer wish to receive these updates.
  */
 export class ContextEvent<T extends UnknownContext> extends Event {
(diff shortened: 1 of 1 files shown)


justinfagnani:
cc @benjamind

proposals/context.md
```diff
@@ -126,40 +129,41 @@ An element which wishes to receive some context and participate in the Context A
...
+      this.myCoolThing = coolThing; // do something with value
+    }
+  )
 );
 ```
 
 If a provider listening for this event can provide the requested context it will invoke the callback passed in the payload of the event. The element can then do whatever it wishes with this value.
 
-It may also be the case that a provider can retain a reference to this callback, and can then invoke the callback multiple times. In this case providers should pass a dispose function as a second argument to the callback to allow consumers to inform the provider that it should no longer update the element, and should dispose of the callback.
+It may also be the case that a provider can retain a reference to this callback, and can then invoke the callback multiple times. In this case providers should pass a unsubscribe function as a second argument to the callback to allow consumers to inform the provider that it should no longer update the element, and should dispose of the callback.
```
Westbrook, line 143:
```suggestion
It may also be the case that a provider can retain a reference to this callback, and can then invoke the callback multiple times. In this case providers should pass an unsubscribe function as a second argument to the callback to allow consumers to inform the provider that it should no longer update the element, and should dispose of the callback.
```

proposals/context.md
```diff
@@ -296,15 +299,15 @@ export type ContextCallback<ValueType> = (
  * A provider should inspect the `context` property of the event to determine if it has a value that can
  * satisfy the request, calling the `callback` with the requested value if so.
  *
- * If the requested context event contains a truthy `multiple` value, then a provider can call the callback
- * multiple times if the value is changed, if this is the case the provider should pass a `dispose`
- * method to the callback which requesters can invoke to indicate they no longer wish to receive these updates.
+ * If the requested context event contains a truthy `subscribe` value, then a provider can call the callback
+ * multiple times if the value is changed, if this is the case the provider should pass a `unsubscribe`
```
Westbrook, line 303:
```suggestion
 * multiple times if the value is changed, if this is the case the provider should pass an `unsubscribe`
```

Westbrook (commented):
LGTM, baring some grammar nits.

@benjamind thoughts?

benjamind:
Yes looks good to me, thanks for pulling this up from the `lit/context` implementation @justinfagnani!

Issue #42, opened 2023-04-26
[context] Event namespace


loganvolkers:
Thanks for putting together a formal specification @justinfagnani

I tried (and failed) at pushing for some centralization some years ago via [dom-context](https://github.com/saasquatch/dom-context)

Two things to share:

1. Please see the below list of related projects. In writing `dom-context` one of the design goals was to be compatible with existing libraries, so I think such an audit would be helpful for moving the proposal forward.
2. Consider making the event name flexible. The spec now considers everything as `context-request`. Other libraries use other event names. It might aid centralization and standardization if we can support polyfills for existing components and libraries (both providers and subscribers).

## Related projects

- [blikblum/wc-context](https://github.com/blikblum/wc-context) - uses the same event handler approach, includes integrations with other Web Component libraries, and is well tested, but doesn't support retries/polling. Uses the `context-request-${name}` event namespace. Exposes a core library, so it can be used in other web component compilers.

- [askbeka/wc-context](https://github.com/askbeka/wc-context) - uses the same event handler approach with the `request-context-${contextName}` namespace. Only works with custom elements, so incompatible with Stencil.

- [petermikitsh/stencil-context](https://github.com/petermikitsh/stencil-context) - uses the same event handler approach, but does not support having different context names (everything uses the same shared `mountConsumer` event name)

- [ionic-team/stencil-state-tunnel](https://github.com/ionic-team/stencil-state-tunnel) - doesn't support nested providers ([see issue #8](https://github.com/ionic-team/stencil-state-tunnel/issues/8#issuecomment-655845289)) and requires javascript props on components to wire them up.

- [mihar-22/stencil-wormhole](https://github.com/mihar-22/stencil-wormhole) - uses the same event handler approach with `openWormhole` and `closeWormhole` event names. Only supports using a single object as context, spreads that object to it's children properties.

- [@corpuscule/context](https://github.com/corpusculejs/corpuscule/tree/master/packages/context) - uses the same event handler approach, but uses decorators so it is incompatible with Stencil

- [haunted](https://github.com/matthewp/haunted) - uses the same event handler approach with `haunted.context` event name, but relies on `detail.Context` objects for handling multiple context types. Only exposes `Provider` as custom HTML elements, so doesn't support global providing, or connecting providers into non-custom elements.

justinfagnani:
[...]

PR #47, opened 2023-11-08
Add a proposal template

matthewp:
Currently there is no suggestion given to the format of a proposal. This has led to proposals that are inconsistent with other and previous proposals. Additionally some proposals have omitted explicit **Goals** which makes it difficult to understand what the proposal is doing and why.

This adds a proposal template to the repo that acts as a starting point. We can discuss which sections should be required or if some should be added.


proposal-template.md
```diff
@@ -0,0 +1,31 @@
+# Proposal name
+
+**Author**: Name here
+
+**Status**: Draft
+
+**Last updated**: Date here
```
Westbrook, line 7: Can you include a created date, as well?

Westbrook (commented):
Great idea!

Would it make sense to sense to include a “Open Questions / Previous considerations” section as per #42?

proposal-template.md
matthewp: Done

README.md
```diff
@@ -6,7 +6,7 @@ Think of things like a context API, async work (Suspense-like), SSR-interop, inc
...
+Check out the [Issues](https://github.com/webcomponents/community-protocols/issues) and [PRs](https://github.com/webcomponents/community-protocols/pulls) currently open on this repo to join in on conversations already inflight regarding a number of exciting areas. If you have ideas on a protocol you like raised to the community open a new [issues](https://github.com/webcomponents/community-protocols/issues/new) so that authors and consumers of libraries and components from across the community that likely have (or wanted to have) implemented features in that area can join in with their use cases as well. Once the rough edges have been hammered out, submit a PR (pull-request) with specs of the protocol and append it to the "Protocols" list below so it can be formalized and put into use across the community. Use this [template](./proposal-template.md) for your proposals PRs.
```
nolanlawson, line 9:
```suggestion
Check out the [Issues](https://github.com/webcomponents/community-protocols/issues) and [PRs](https://github.com/webcomponents/community-protocols/pulls) currently open on this repo to join in on conversations already inflight regarding a number of exciting areas. If you have ideas on a protocol you'd like raised to the community, open a new [issue](https://github.com/webcomponents/community-protocols/issues/new) so that authors and consumers of libraries and components from across the community who likely have (or wanted to have) implemented features in that area can join in with their use cases as well. Once the rough edges have been hammered out, submit a PR (pull-request) with specs of the protocol and append it to the "Protocols" list below so it can be formalized and put into use across the community. Use this [template](./proposal-template.md) for your proposal's PRs.
```

Just some minor copyedits.

justinfagnani (approved):

mattlucock:
@matthewp Sorry to bring up an issue just after this has been merged, but I'm actively contemplating using the template, and I'm a little unsure about the ordering of the sections. Having the examples section immediately follow the summary section at the beginning of a proposal feels strange to me, since it seems like the examples are necessarily going to illustrate concepts that have not yet been introduced or explained, which seems like it would be quite confusing for the reader. What was your rationale for having it at the beginning? I feel like the examples would most logically follow after the design detail section.

matthewp:
@mattlucock I mostly copied from RFC templates in existing OS projects. I think the ordering here makes sense, the `Design detail` section can be pretty verbose, maybe even being a step-by-step of how the thing should be implemented. Seeing such details before you have an idea of the shape of the API can be difficult. For me at least, seeing an example is enough to understand what the API is doing. The details then explain, well, the details.

Of course the template is just a template and if you feel your proposal should be ordered differently you are able to do so. It's not a rule that it be ordered this way.

PR #59, opened 2024-04-04
[context] Fix type inconsistencies.

### LINKED ISSUE #58: [context] Type definitions aren't consistent.
It looks like I missed updating some of the type declarations in the protocol description in https://github.com/webcomponents-cg/community-protocols/pull/36

This has caused some [interop problems](https://github.com/lit/lit/issues/4609) with implementations that used the previous object-based contexts that we decided to change in https://github.com/webcomponents-cg/community-protocols/issues/19.

justinfagnani:
Fixes #58

As mentioned in https://github.com/webcomponents-cg/community-protocols/issues/58, I missed some type definitions when fixing #19. This fixes those.

I added one small rename for clarity, of `ContextEvent` -> `ContextRequestEvent`. This is more explicit and leaves room for other context related events, like maybe a `ContextProviderEvent` as a possible solution to https://github.com/webcomponents-cg/community-protocols/issues/25

The interface name _shouldn't_ be too significant here, since we aren't publishing these types, _but_:
1. People are copying the types from this README.
2. We _should_ publish the types here so that implementor can depend on them and get help from the TypeScript compiler to check that their implementation is at least type-compatible with the protocol. (we should also create a conformance test suite).

I copied these fixed types into `@lit/context` tests and discovered one place where the Lit types were incorrect, so they would be very useful for that sort of check.

### DIFF
--- a/proposals/context.md
+++ b/proposals/context.md
@@ -96,11 +96,11 @@ Then values can be cast to this to create a typed context key:
 export const myContext = 'my-context' as Context<string, number>;
 ```
 
-The type of a Context can then be extracted with a utility type:
+The value type of a Context can then be extracted with a utility type:
 
 ```ts
-export type ContextType<Key extends Context<unknown, unknown>> =
-  Key extends Context<unknown, infer ValueType> ? ValueType : never;
+export type ContextType<T extends UnknownContext> =
+  T extends Context<infer _, infer V> ? V : never;
 ```
 
 Usage:


Westbrook (approved):
LGTM! Not a specifically functional change, so I don't see any issue with merging it straight away, but want to make space for "the community" to get a chance to look at this before doing so.

proposals/context.md
```diff
@@ -272,37 +272,30 @@ Below are some TypeScript definitions for the common parts of the proposed proto
...
-  name: string,
-  initialValue?: T
-): Readonly<Context<T>> {
-  return {
-    name,
-    initialValue,
-  };
-}
+export const createContext = <ValueType>(key: unknown) =>
+  key as Context<typeof key, ValueType>;
```
augustjk, line 298: I have a question about typings here but since this PR is just fixing inconsistencies, I raised it in a separate issue https://github.com/webcomponents-cg/community-protocols/issues/60

proposals/context.md
```diff
@@ -272,37 +272,30 @@ Below are some TypeScript definitions for the common parts of the proposed proto
...
+export type UnknownContext = Context<unknown, unknown>;
 
 /**
  * A helper type which can extract a Context value type from a Context type
  */
-export type ContextType<T extends UnknownContext> = T extends Context<infer Y>
-  ? Y
-  : never;
+export type ContextType<T extends UnknownContext> =
+  T extends Context<infer _, infer V> ? V : never;
```
augustjk, line 292: This looks different from what's above
https://github.com/webcomponents-cg/community-protocols/blob/952f15841a809fec73bbaeed842809a2d88b4deb/proposals/context.md?plain=1#L101-L104

augustjk (commented):
Should the "Last update" at the top of the doc also be updated?

PR #66, opened 2024-11-11
Adding simple slottable request example

JimSchofield:
I thought it would be helpful to show a very simple example of the request working so people could grok the simple case before looking at the more complicated examples provided.


Westbrook (approved):
LGTM. Thanks for giving us some coverage at the low end here!

PR #73, opened 2025-04-22
Proposal: Common interface for ElementInternals

davatron5000:
There's no standard API to expose `ElementInternals` to `@public`. Third-party tooling like Automated Accessibility Testing tools are unable to reliably "guess" and find those exposed internals (e.g. cannot find `aria-role`). It would be good for the Web Components Community Group to have an agreed upon standard for exposing `ElementInternals` so other tooling can build from our convention.

### DIFF
--- a/proposals/element-internals-declaration.md
+++ b/proposals/element-internals-declaration.md
@@ -1,0 +1,74 @@
+# Proposal name
+
+**Author**: Dave Rupert
+
+**Status**: Proposal
+
+**Created**: April 22, 2025
+
+**Last updated**: April 22, 2025
+
+# Summary
+
+There's no standard API to expose `ElementInternals` to `@public`. Third-party tooling like Automated Accessibility Testing tools are unable to reliably "guess" and find those exposed internals (e.g. cannot find `aria-role`). Problems that arise from this are best summed up on the DequeuLabs aXe-core Github Repo:
+
+https://github.com/dequelabs/axe-core/issues/4259
+
+Both aXe (on thread) and Microsoft Accessibility Insights (privately) have indicated that a community standard here would go a long way to supporting the probing of `ElementInternals` for valid roles.
+
+Another potential win here would be an SSR story where tooling to make static HTML could elevate internals like `role` or a default `:state()` to the template.
+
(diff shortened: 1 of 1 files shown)


sorvell:
What about just using `attachInternals()` directly?

```js
// convention
#internals: ElementInternals;
attachInternals() {
  return (this.#internals ??= super.attachInternals());
}
```

jonathantneal:
I like this proposal, @sorvell, because it makes `internals` an introspectable property.

The optionality here is powerful — it lets us check for an `internals` property without necessarily creating `ElementInternals`. With the pattern you suggested, the only way to know whether a component uses `ElementInternals` is to create it if it doesn’t already exist. This proposal avoids that by allowing internals to be checked (akin to `shadowRoot`) so downstream code can detect and interact with it only if it’s already in use.

jonathantneal:
Separately, I want to add that I do like the pattern you shared, @sorvell, independent of this proposal.

I think there’s nothing stopping an author from creating a mixin or base element that implements exactly what you describe while still fulfilling the shape of this proposal:
```ts
class extends HTMLElement {
	internals: ElementInternals = super.attachInternals()

	attachInternals(): ElementInternals {
		return this.internals
	}
}
```

justinfagnani:
I don't think we should change the semantics of `attachInternals()`. You can allow mixins to call it while preserving the call-only-once behavior.

```ts
#internals: ElementInternals;
#attachInternalsCalled = false;

constructor() {
  super();
  this.#internals = super.attachInternals();
}

attachInternals() {
  if (this.#attachInternalsCalled) {
    throw new Error(standardErrorMessage);
  }
  this.#attachInternalsCalled = true;
  return this.#internals;
}
```

justinfagnani:
One way to expose internals in testing environments is to look for a global WeakMap to add internals to. The name of the global would be the protocol:

```ts
constructor() {
  super();
  this.#internals = this.attachInternals();
  window.__elementInternalsForTesting?.set(this, this.#internals);
}
```

Then a test environment can just do:
```ts
window.__elementInternalsForTesting = new WeakMap();
```

PR #74, opened 2025-04-22
[context] Use stopImmediatePropagation()

jonathantneal:
This PR updates the Context Protocol specification to use `stopImmediatePropagation()` instead of `stopPropagation()`. This change ensures that event handling is more precise and prevents potential issues where multiple providers might try to handle the same context request.

The change is important because:
- `stopImmediatePropagation()` prevents other event listeners on the same element from being called.
- This provides better control and clarity over which provider handles a context request.
- It prevents potential race conditions where multiple providers might try to satisfy the same request.

The specification already had consistent usage of stopping propagation throughout, so no other changes were needed.

Additionally, if it seems helpful, I can add a note to the proposal explaining the rationale behind `stopImmediatePropagation`, such as:

> Using `stopImmediatePropagation` maintains a clear contract that once a provider commits to handling a request, it alone is responsible for satisfying it, and no other providers should intervene. This ensures that no other event listeners on the same node — or any node further up the tree — can handle the same context request, avoiding race conditions where multiple providers might attempt to respond.

### DIFF
--- a/proposals/context.md
+++ b/proposals/context.md
@@ -43,7 +43,7 @@ At a high level, the Context API is an event-based protocol that components can 
 - A component requiring some data fires a `context-request` event.
 - The event carries a `context` value that denotes the data requested, and a `callback` which will receive the data.
 - Providers can attach event listeners for `context-request` events to handle them and provide the requested data.
-- Once a provider satisfies a request it calls `stopPropagation()` on the event.
+- Once a provider satisfies a request it calls `stopImmediatePropagation()` on the event.
 
 # Details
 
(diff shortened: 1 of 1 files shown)


justinfagnani:
@Westbrook @benjamind this change looks great, but it highlights the fact that we don't have anything specified for protocol change management, outside of the "last updated" date.

Should we add a small changelog to the end of the protocol readme to track changes?

justinfagnani:
@jonathantneal this looks good! You need to update the last changed date a the top of the doc, and 
depending on what we think about my previous comment, maybe a Changes section at the end. Thanks!

PR #75, opened 2026-04-23
Exposing ElementInternals via global WeakMap

bennypowers:
Add examples for global weakmap and well-known properties related to element internals.

This follows up on discussions held in the [axe core github](https://github.com/dequelabs/axe-core/issues/4259#issuecomment-2885569017) as well as face-to-face calls between Adobe, Deque, and Red Hat

This PR revises @davatron5000 's original in a few main ways:
- replaces the public `this.internals` field on elements with `globalThis._elementInternals ??= new WeakMap()`
- removes mention of "publicizing" internals at all
- focuses the exposure of internals on accessibility audit tools

If this PR is well written, then the proposal should be understood to:
- not require any public API on elements
- clarify that this protocol is a _stopgap_ while browser vendors work out a better solution to expose the ax tree to tools
- be sorely needed by real teams today


Westbrook (commented):
Is this a "maybe we could do this?" or a "axe will support this!" or a something else?

If it's support, is there a member of their team to approve/provide a link we could point to as a "I know, crazy, but it's true...see this page"?

bennypowers:
It's the second one. Axe will support the global weakmap, as well as a variety of element dom props they've seen in the wild.

That's correct for a tool like axe, which needs to cast a wide net to validate real-world web pages. I'd prefer that the community protocol recommend the global weakmap only, though, as that would be least disruptive to element APIs

 NB ultimately this is a bad hack, necessary because vendors haven't exposed ax tree apis for this kind of thing. We should really want to see that, but this protocol is necessary in the mean time

bennypowers:
I've updated my PR based on feedback:

- There is no goal to make internals public. They are called "ElementInternals" for a reason
- Clarify that this is a stopgap until browser vendors ship better ax tree APIs
- Remove public property API (axe core will support this anyways, but we shouldn't recommend it, since that adds public tech debt for a stopgap)

bennypowers:
after this is reviewed, revised as needed, and merged, it should be referenced in `https://github.com/w3c/aria/issues/2663` - which i put here in backticks specifically to avoid backreferences, until these changes (and their revisions) are accepted

Westbrook (approved):
LGTM!

Thanks for keeping this conversation going with the aXe team and for bringing the findings back.

PR #76, opened 2026-08-13
[context] Propose a native API for the Context Protocol

jonathantneal:
This PR revises the **Context Protocol proposal** to clarify its behavior and define a complete lifecycle for operations previously left to individual libraries:

- Provider lookup still follows the composed DOM event path, but the nearest matching provider now responds explicitly using `event.respond()`.
- Consumers can await one value with `element.requestContext()`, which returns a promise that remains pending if no provider is yet available.
- Consumers can receive ongoing updates with `element.subscribeContext()` until its signal is aborted.
- The browser manages the operation lifecycle, including pending requests, cancellation, announced provider changes, subscription handoff, document adoption, and teardown.

The result preserves the protocol’s decentralized, shadow-DOM-friendly provider model while giving authors a simpler consumer API and implementations a consistent, testable lifecycle.

#### Example: Direct dispatch

A consumer dispatches a context request, and the nearest matching provider responds along the event path:

```js
const themeContext = Symbol('theme');
const provider = document.querySelector('theme-provider');
const consumer = provider.querySelector('profile-card');

provider.addEventListener('context-request', (event) => {
  if (event.context === themeContext) {
    event.respond(defaultTheme);
  }
});

consumer.dispatchEvent(
  new ContextRequestEvent('context-request', {
    context: themeContext,
    callback(theme) {
      consumer.theme = theme;
    },
    bubbles: true,
    composed: true,
  })
);
```
> This preserves synchronous, event-based lookup while giving providers an explicit response contract instead of requiring them to invoke a consumer callback directly.

#### Example: Awaiting one value

A component can request a value from its nearest matching provider. If no provider is currently available, the promise remains pending until one responds:

```js
const themeContext = Symbol('theme');
const profileCard = document.querySelector('profile-card');

const theme = await profileCard.requestContext(themeContext);
```
> This fills the “no provider yet” gap: a one-time request can remain pending and resolve when a matching provider later becomes available.

#### Example: Receiving updates

A consumer can subscribe to values:

```js
const themeContext = Symbol('theme');
const profileCard = document.querySelector('profile-card');
const controller = new AbortController();

profileCard.subscribeContext(
  themeContext,
  (theme) => {
    profileCard.theme = theme;
  },
  { signal: controller.signal }
);

// ... at some point later stop receiving updates.
controller.abort();
```
> This standardizes subscription lifetime and cancellation, while allowing the browser to manage provider replacement and unsubscribe handoff.

## Feedback requested

If this direction is appropriate, I would appreciate feedback on:
- consumer methods versus provider events as the API boundary;
- `contextTarget` and lookup across shadow boundaries;
- cancellation and subscription lifetimes; and
- migration for existing event-only implementations.

## How to read this

This is a near-complete rewrite, so please read the [rendered proposal](https://github.com/jonathantneal/community-protocols/blob/jon/context-protocol-finalization/proposals/context.md) rather than the raw diff.

The opening API overview and [Usage](https://github.com/jonathantneal/community-protocols/blob/jon/context-protocol-finalization/proposals/context.md#usage) examples provide the quickest introduction. The detailed processing model, rationale, and proposed web-platform test coverage follow.

## Compatibility

Existing event-only consumers and providers continue interoperating with one another unchanged. Mixed native and event-only participants require the documented dual-mode provider path, which feature-detects `respond()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Check out the [Issues](https://github.com/webcomponents/community-protocols/issu
 
 | Proposal       | Author           | Status     |
 |----------------|------------------|------------|
-| [Context]      | Benjamin Delarre | Candidate  |
+| [Context]      | Benjamin Delarre | Draft      |
 | [Defer Hydration] | Justin Fagnani | Proposal |
 | [Pending Task] | Justin Fagnani   | Draft      |
 | [Slottable Request] | Kevin Schaaf   | Proposal  |
--- a/proposals/context.md
+++ b/proposals/context.md
@@ -1,357 +1,872 @@
 # Context Protocol
 
-An open protocol for data passing between components.
+An open protocol for passing contextual data between components.
 
 Author: Benjamin Delarre
 
 Document status: Candidate
 
-Last update: 2025-05-02
+Last update: 2026-08-13
 
 # Background
 
-There are a number of scenarios where a web component might require data that is provided from outside itself. While components can specify properties and attributes to receive that data imperatively or declaratively, it is often the case that the data may be owned somewhere further up the DOM tree.
-
-One approach to passing this data down to components is commonly referred to as *prop drilling*, whereby components pass properties all the way down through the hierarchy, passing from component to component until it is consumed at its destination. This is generally considered undesirable as it often requires intermediate components in the tree to have knowledge of the data necessary in its descendents.
-
-Frameworks and libraries often provide mechanisms for this, these can range from the simple Context implementation available in React, to more complex Dependency Injection frameworks. Web components need a similar protocol in order to solve this problem.
+Web components often need data owned by an ancestor, such as a theme, locale, logger, or state store.
+Passing that data through every intermediate component couples those components to data they do not otherwise use.
+Frameworks commonly solve this with context or dependency-injection APIs, but components built with different libraries need a shared browser API.
 
 # Goals
 
-- Allow elements in the DOM to retrieve data based on their contextual position in the DOM
-- Alleviate the problem of *prop drilling*
-- Simple API that is easily implemented in any framework / library
-- Synchronous protocol, while supporting asynchronous patterns
-- Support single or multiple delivery of context values
+- Resolve context from the nearest matching provider on an element's composed event path.
+- Avoid passing contextual data through unrelated intermediate components.
+- Provide synchronous discovery when a provider is already available.
+- Let one-time requests wait for a later provider.
+- Support subscriptions, provider replacement, and explicit cancellation.
+- Give frameworks and plain custom elements the same API.
 
-# Non-Goals
+# Non-goals
 
-**Context API !== Dependency Injection Framework**
+The Context Protocol is not a dependency-injection framework.
+It does not define constructor, factory, or property injection.
+A dependency-injection system may use this protocol to locate dependencies, but those higher-level patterns are outside its scope.
 
-The Context API does not intend to cover all cases and forms of Dependency Injection. It does not specify constructor, factory or property injection patterns. Its only goal is to formalize the pattern of sharing data across the hierarchy in the DOM, specifically avoiding *prop drilling* type scenarios. Dependency Injection patterns could be implemented using this protocol, but this is not the goal and should remain explicitly outside the scope of Context API for simplicity.
+The Context Protocol is also not a state-management system.
+A component may use it to obtain a state store, but the protocol does not define how state is stored, updated, or observed.
 
-**Context API is not a state management alternative**
+# API
 
-State management libraries often need to solve some similar problems that the Context API helps to solve. An element deep in the DOM tree may need access to some state, and may need to respond to that state being changed.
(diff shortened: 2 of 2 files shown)


JSteitz:
## Feedback

I wonder whether some subscription logic could be substantially simplified.

The common theme behind the points below is that the protocol currently exposes
and maintains several concepts — `signal`, delivery callback, provider
registration, unsubscribe callback, and operation state — where I think the
provider could instead be given a single response capability whose lifetime is
managed by the UA.

### 1. Exposing `signal`

I don't think the provider needs access to the `AbortSignal` itself.

The provider only needs to know whether the request is still active. Exposing a
`readonly aborted` attribute would provide that information without exposing
the full `AbortSignal` API.

For example:

```js
if (event.aborted) {
    return;
}

event.respond(value);
```

This seems like a simpler provider-facing API while still allowing the UA to
use an `AbortSignal` internally for managing operation cancellation.

I don't see a need for the provider to observe the `"abort"` event itself. If
there is a use case that requires asynchronous notification of cancellation, it
would be useful to understand what that use case is.

### 2. I would not expose the delivery callback

I don't see a strong reason for the provider to receive a separate
`event.callback`.

The current model makes `respond()` establish/confirm the provider registration
and then uses a separate delivery callback for subsequent subscription updates.
This means the provider has to retain another capability and the UA has to
maintain the relationship between that callback, the operation, the current
provider, and its unsubscribe function.

Could `respond()` itself be the capability used for both the initial response
and subsequent subscription updates?

In other words, instead of having one capability for the initial response and
another callback for subsequent delivery, I would prefer something conceptually
closer to:

```js
event.respond(value, unsubscribe);

// later
event.respond(value);
```

where the response capability remains valid for the lifetime of the provider's
active subscription.

The important security checks would still belong in the UA. `respond()` could
verify that the capability is still associated with the current provider and an
active operation before delivering the value. This would avoid giving the
provider a second delivery capability whose lifecycle and validity have to be
managed separately.

### 3. Simplify subscription handling

This follows from the previous point.

If `respond()` were reusable for subscriptions, I think the subscription
handling could be considerably simpler.

For example, a provider could retain the request events for its active
subscriptions:

```js
const subscribers = new Set();

this.addEventListener('context-request', event => {
    if (event.context !== 'my-context') {
        return;
    }

    // Guard against repeated one-time requests.
    if (event.responded) {
        return;
    }

    // Guard against requests whose operation has been aborted.
    if (event.aborted) {
        return;
    }

    // Register this request as an active subscription
    if (event.subscribe && !subscribers.has(event)) {
        subscribers.add(event);

        event.respond('some_value', () => {
            subscribers.delete(event);
        });

        return;
    }

    // Respond as one-time request (no subscription support)
    event.respond('some_value');
});

function updateContextListeners() {
    for (const event of subscribers) {
        event.respond('some_updated_value');
    }
}
```

The essential difference from the current proposal is that there is **no
separate delivery callback**. The provider retains the request's response
capability and calls `respond()` again when the context value changes.

The exact implementation could, of course, differ, but the key part is that
`respond()` itself becomes reusable for an active subscription.

The UA would still own the operation, provider handoff, abort, and validity of
the response capability. The `Set` is simply provider-side bookkeeping for the
active subscriptions.

This also seems to make the distinction between one-time requests and
subscriptions straightforward:

- for a one-time request, `respond()` can only successfully claim the request
  once;
- for a subscription, the provider that owns the subscription can continue
  calling `respond()` to deliver updates.

I think this could remove a significant amount of subscription-specific
complexity while retaining the UA's ability to enforce the relevant security
and lifecycle invariants.

### 4. Pending one-time requests

I'm not sure what use cases the pending state for one-time requests is intended
to solve.

What concrete scenarios require a one-time context request to remain pending
rather than simply being dispatched again when the requester knows that its
context may have changed?

And, more importantly, are those scenarios actually requirements that this
protocol needs to solve?

I can see that pending requests can be useful when a provider is not yet
available, but I'm not sure whether the protocol needs to retain arbitrary
one-time requests and re-evaluate them as part of its operation model.

I'd like to better understand the intended use cases and why they need to be
handled by the protocol itself.

### 5. Native `requestContext()` / `subscribeContext()` methods

I don't see much benefit in adding the native `requestContext()` and
`subscribeContext()` methods to `Element`.

As I understand it, the provider still has to listen for `context-request` and
handle the event accordingly, so these methods appear to primarily be
convenience helpers for dispatching the corresponding requests.

If they are intended to provide semantics beyond convenience wrappers around
the event protocol, it would be useful to clarify what those semantics are.

Otherwise, I wonder whether these methods are necessary as part of the protocol
itself, rather than being something that could be provided by a small
helper/library API.

## Overall

The protocol already has a fairly sophisticated UA-owned operation model for
cancellation, provider handoff, re-evaluation, document ownership, adoption,
and garbage collection. I think that can remain UA-internal while presenting a
much smaller provider-facing API.

In particular, the UA can continue to control the validity and ownership of the
response capability. This could eliminate the need for a separate
provider-facing delivery callback and simplify the subscription model without
giving up the lifecycle and security guarantees provided by the operation
model.

